PR #12, opened 2021-04-01
Merge of VL and Jotunn localization code

Algorithman:

PR #15, opened 2021-04-01
Feat/algorithman/patch initializer rework

Algorithman:
Reworked the patch initialization system

New PatchInit attribute, used on public static methods where your On.xxxx hooks should reside.
Will be called right after plugin initialization, in order of their priority value which allows a bit of flexibility.
Multiple PatchInit methods can be set within a class, each with their own priority.


RatikKapoor:
Depends on #12

raduschirliu (approved):

RatikKapoor (approved):

PR #20, opened 2021-04-02
Feat/algorithman/console command split quotes

Algorithman:
Proper splitting of argument string for console commands, quotes are taken into account as well.


Algorithman:
was already open chain

Algorithman:
Recreating this one from dev

PR #24, opened 2021-04-03
fix: Patch initialization call was lost in merge

Algorithman:
The call of the InitializePatches method was lost in merging


paddywaan (approved):

sirskunkalot (changes requested):
I will add it in my branch. I renamed JotunnLib.cs. So it will have the same not resolvable conflicts as last time...

sirskunkalot:
incorporated into #31

PR #32, opened 2021-04-03
Unit test

MSchmoecker:
Disclaimer: I made only a few unit test on small projects so far, no with unity modding especially. 

I have setup a unit test project for Jotunn, this should be useful in the long run. It uses basic NUnit, not the Unity Test Framework. A few changes had to be made, I removed MonoBehaviour from Manger.cs to properly test these child classes. This should be not be a problem because those managers don't use MonoBehaviour as far I've seen.
I added a few tests to the ObjectManager class as a demonstration (JotunnTest/ObjectManagerTests.cs).

This is only a proposal, it's up to you guys if you want to write unit tests along. I do not intent to maintain this, only write a few test cases.

### DIFF
--- a/JotunnLib.sln
+++ b/JotunnLib.sln
@@ -26,6 +26,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "resources", "resources", "{
 		README.md = README.md
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JotunnTest", "JotunnTest\JotunnTest.csproj", "{BF1FE878-2FA2-4CAE-9C3B-0966BC7CF8EE}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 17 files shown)


sirskunkalot (changes requested):
Hi MSchmoecker,

we are in the middle of a merger between ValheimLib and JotunnLib. There was a big change pending which got approved now so I must ask you to merge the current dev into your PR and then resubmit. 

And may I ask upfront: Why did you delete the MonoBehaviour in the Manager class? Some managers rely on the Unity messages being sent automatically on certain events. So that would be a great stopper, i guess. Is it possible to achieve the Unit tests kepping the MonoBehaviour base?

Regards,
Jules

MSchmoecker:
It is not possible with MonoBehaviour because a class that inherits it cannot be instanitiated outside of Unity. So you cannot call `new ObjectManager()` for example. And therefore all manager classes cannot be tested.
The only way to keep MonoBehaviour is to use the [Unity Test Framework](https://docs.unity3d.com/2019.4/Documentation/Manual/testing-editortestsrunner.html) and having a Unity project that inlcudes all necessary dll files.

paddywaan:
So, I'm not entirely happy with our implementation of singleton and was thinking about implementing a more generic DI solution which should also make things a little more simple for unit testing, however i'll need to talk with everyone else. If you want to help us contribute to this project and be a little more involved in discussions please add me on discord and i'll add you to the group chat. Thank's for your PR and we'll have a think about how we will resolve some of these issues.

paddy#1337

MSchmoecker:
Thanks for your response, I'm closing the PR now.

PR #42, opened 2021-04-04
Fix savemanager lag

xiaoxiao921:


sirskunkalot (approved):
sure, all in for some runtime IL patching

paddywaan (approved):

PR #46, opened 2021-04-05
feat: Primary checks for installed mods/mod versions

Algorithman:
feat: Primary checks for installed mods/mod versions
TODO: Text output on the shown panel if version/mod mismatch is observed

Test case now works off a config value in the JotunnLib configuration (debug builds only)

---- obsolete ----
To check functionality, remove the #if TESTCASE in ModCompatibility.cs, line 141. This will let the server return a wrong version string. 
---- obsolete ----


sirskunkalot (approved):
works so far, server and local :)

paddywaan (approved):

PR #66, opened 2021-04-08
Feat/algorithman/gui manager additions

Algorithman:
feat: network compatibility check and error message
- added more granularity to the check (thx @paddywaan)
- moved the check in front of RPC_PeerInfo
- client version data is appended now to the original ZPackage (so there should be no impact for clients without JotunnLib), no more hooking of Version.GetVersionString()
- added GUIManager functions
- added valheim style error message window
- added sanity checks for items and pieces (can't load two items of the same name)
- added lulz


sirskunkalot (approved):
for the lulz

paddywaan (approved):

PR #67, opened 2021-04-08
fix: hidden 2 methods which should not be exposed, fixed capitalization

Algorithman:
just a little checkup for exposed methods

### DIFF
--- a/JotunnLib/Managers/ZoneManager.cs
+++ b/JotunnLib/Managers/ZoneManager.cs
@@ -12,7 +12,7 @@ namespace JotunnLib.Managers
 
         internal override void Init()
         {
-            On.ZNetScene.Awake += registerAllToZNetScene;
+            On.ZNetScene.Awake += RegisterAllToZNetScene;
         }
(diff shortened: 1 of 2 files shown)


sirskunkalot (approved):

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -26,7 +26,7 @@ private void Awake()
             Instance = this;
         }
 
-        public void registerAllToZNetScene(On.ZNetScene.orig_Awake orig, ZNetScene self)
+        private void RegisterAllToZNetScene(On.ZNetScene.orig_Awake orig, ZNetScene self)
```
raduschirliu, line 29: Could we rename this to be lowerCamelCase if it's private? Or make it internal and leave the casing?

raduschirliu: Nvm, my bad.

raduschirliu (approved):

PR #73, opened 2021-04-08
fix: Modded items were being saved into wrong folder

raduschirliu:
fix: Modded items were being saved into "Main" folder instead of correct folder. Changed to use library name for save dir instead.
This should be merged after #72, since it seems one of the changes from there made its way into this branch :(


sirskunkalot (approved):
lets see if this builds again

Algorithman (approved):

RatikKapoor (approved):

PR #91, opened 2021-04-12
Feat/algorithman/localization manager changes

Algorithman:


JotunnLib/Managers/LocalizationManager.cs
```diff
@@ -228,32 +228,28 @@ private void AddLanguageFilesFromPluginFolder()
         /// <param name="forceReplace">replace the token if it already exists</param>
         public void AddToken(string token, string value, string language = DefaultLanguage, bool forceReplace = false)
         {
-            if (token[0] != TokenFirstChar)
```
sirskunkalot, line 231: why did you strip that? for the tokens it should be fine.

paddywaan (approved):

raduschirliu (approved):

JotunnLib/Managers/LocalizationManager.cs
raduschirliu: Wouldn't we want to treat single tokens in the same way that we treat an entire localization dictionary though?
If we can add a dictionary that doesn't require tokens to start with `$`, I would think that adding single tokens should behave the same way

sirskunkalot: OK, if you are sure of that change, lets add it :)

raduschirliu, line 231: Would we be able to get rid of `TokenFirstChar` entirely now? Or is it still being used somewhere

Algorithman: Still used in Skillsconfig iirc. Search for "$" (including the quotes)

PR #98, opened 2021-04-13
Make prebuild task optional

Algorithman:
- moved JotunnLib props files into BuildProps folder
- removed duplicate calls for the build task
- added DoPrebuild.props, with the ExecutePrebuild defined inside (**true**(default)/false). Prebuild task will only be called if true. 
**Be sure not to push DoPrebuild.props to the server!**
Either add it to your local excludes or revert the file before pushing.


paddywaan:
Should we not just include it in our gitignore instead of local gitignore then?

sirskunkalot (approved):
whoopsie

paddywaan (approved):

PR #104, opened 2021-04-14
fix: InputManager now can bind multiple mods to 'one' button

Algorithman:
fix: InputManager now can bind multiple mods to one button (internally it's two buttons). Downside until v1.1, both mods trigger also on each other's key.


paddywaan (approved):

sirskunkalot (approved):
fair enough

PR #110, opened 2021-04-15
fix: resolved issue which caused mock references with (instance) suff…

paddywaan:
…ix to fail to resolve due to lack of capital i(I).

### DIFF
--- a/JotunnLib/Extensions/MockExtentions.cs
+++ b/JotunnLib/Extensions/MockExtentions.cs
@@ -84,7 +84,7 @@ namespace JotunnLib
                     // Cut off the suffix in the name to correctly query the original material
                     if (unityObject is Material)
                     {
-                        const string materialInstance = " (instance)";
+                        const string materialInstance = " (Instance)";
                         if (unityObjectName.EndsWith(materialInstance))
                         {
                             unityObjectName = unityObjectName.Substring(0, unityObjectName.Length - materialInstance.Length);


xiaoxiao921 (approved):

JotunnLib/Extensions/MockExtentions.cs
```diff
@@ -84,7 +84,7 @@ public static Object GetRealPrefabFromMock(Object unityObject, Type mockObjectTy
                     // Cut off the suffix in the name to correctly query the original material
                     if (unityObject is Material)
                     {
-                        const string materialInstance = " (instance)";
+                        const string materialInstance = " (Instance)";
```
sirskunkalot, line 87: better to be compared case insensitive.

sirskunkalot (changes requested):

JotunnLib/Extensions/MockExtentions.cs
paddywaan: Not in the case where this is running recursively. The instance string is automated by unity's clone process. The string is never not " Instance". All we are doing here is wasting cpu cycles in a O(n!) task with nothing to be gained. Lets not do that. Its not like I didn't consider proper practice for string comparisons using ordinalignorecase.

sirskunkalot: fair, approved

sirskunkalot (approved):

PR #143, opened 2021-04-20
fix: closes #141

### LINKED ISSUE #141: Skill localizations not correct for custom skills
When custom skills are levelled up, the localizations are not correct on the message displayed on the screen.

The `RaiseSkill` function displays a message like so:
```cs
this.m_player.Message((int) level == 0 ? MessageHud.MessageType.Center : MessageHud.MessageType.TopLeft, "$msg_skillup $skill_" + skill.m_info.m_skill.ToString().ToLower() + ": " + (object) (int) skill.m_level, 0, skill.m_info.m_icon);
```
which attempts to use the localization for the skill's ID

image

paddywaan:
Queries our own collection instead of the games, or returns default.

### DIFF
--- a/TestMod/TestMod.cs
+++ b/TestMod/TestMod.cs
@@ -39,7 +39,7 @@ namespace TestMod
         private ButtonConfig evilSwordSpecial;
         private CustomStatusEffect evilSwordEffect;
 
-        private static Skills.SkillType cType;
+        private Skills.SkillType testSkill;
 
         // Load, create and init your custom mod stuff
         private void Awake()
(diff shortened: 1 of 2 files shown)


TestMod/TestMod.cs
```diff
@@ -272,7 +285,7 @@ private void AddSkills()
             });
 
             // Test adding skills from JSON
-            SkillManager.Instance.AddSkillsFromJson("TestMod/Assets/skills.json");
+            //SkillManager.Instance.AddSkillsFromJson("TestMod/Assets/skills.json");
```
raduschirliu: Shouldn't we leave this in to test the JSON skills as well?

raduschirliu (approved):

TestMod/TestMod.cs
paddywaan: my bad, nicely spotted

sirskunkalot (approved):

PR #155, opened 2021-04-23
Fix/paddy/fix manifest

paddywaan:


manifest.json
```diff
@@ -1,9 +1,10 @@
 {
-    "name": "JotunnLib",
-    "version_number": "0.1.4",
+    "name": "Jotunn, the Valheim Library",
```
sirskunkalot, line 2: dont want to test thunderstores utf compat? ;)

manifest.json
```diff
@@ -1,9 +1,10 @@
...
+    "version_number": "1.0.0",
     "website_url": "https://github.com/jotunnlib/jotunnlib",
-    "description": "JotunnLib is a modding library for Valheim, with the goal of making the lives of mod developers easier. It's a centralized API to allow mod developers to easily add custom items, recipes, prefabs, and more to the game, working for both servers and clients.",
+    "description": "Jötunn, the Valheim Library was created with the intent to facilitate developer creativity, unify the communities problem solving efforts, and enhance developer productivity by curating a library of common helper utilities, as well as interfaces and abstractions which aid with mod interoperability and networked gameplay consistency.",
     "dependencies": [
-        "denikson-BepInExPack_Valheim-5.4.800"
+        "denikson-BepInExPack_Valheim-5.4.901",
+		"ValheimModding-HookGenPatcher-0.0.2"
```
sirskunkalot, line 8: is it a dep to the bepinexpack?

paddywaan: Yes. This enables the modmanagers to acquire dependencies automatically

sirskunkalot: really? you dont say :D but seriously, is hookgenpatcher mandatory for bepinexpack? cause we ship our own if the user enables it...

paddywaan: What? I don't understand what you are asking. Hookgenpatcher is mandatory for us. We use it. We also use bepinex. Have you not released onto thunderstore before?

sirskunkalot: bepinex is an external dep for jötunn, right. but hookgen isnt. the nuget has its own hookgen which can be activated with doprebuild. or is thunderstore plain dll download only?

sirskunkalot: no, dont even have an account there.

paddywaan: Yes, but players dont use nugets. players use dll's. You are looking at this with the wrong perspective.

PR #180, opened 2021-04-28
fix: resolved poor instructions in quickstart regarding MMHook depend…

paddywaan:
…ency for manual jotunn installation


sirskunkalot (changes requested):
file has moved. pls check if still valid

sirskunkalot:
wontfix it seems

PR #184, opened 2021-05-01
prevent crash when plugin info instance is null

tiorthan:
Ran into a situation where Instance was null. I have to look into what caused this later, but I guess this should be checked anyway.


Algorithman (approved):

sirskunkalot (approved):
good catch. totally forgot that one.

PR #196, opened 2021-05-04
Counter early Localization init from mods

### LINKED ISSUE #193: Dont access Localization.instance anywhere before the hooks
First mod to access Localization.instance in it's Awake() will trigger the singleton init which calls LoadLanguage and SetupLanguage. All mods loaded after the mod which first accessed the singleton who rely on hooks to Load/SetupLanguage dont get their hooks called then.

From Localization.cs
grafik

Usage analyze shows there is no other call in code except the singleton init:
grafik

sirskunkalot:
closes #193

### DIFF
--- a/JotunnLib/Managers/LocalizationManager.cs
+++ b/JotunnLib/Managers/LocalizationManager.cs
@@ -215,7 +215,8 @@ namespace Jotunn.Managers
 
             if (token.Any(x => LocalizationEndChars.Contains(x)))
             {
-                throw new Exception($"Token has an end char in it ({LocalizationEndChars}).");
+                Logger.LogError($"Token '{token}' must not have an end char in it ({LocalizationEndChars}).");
+                return;
             }
 
             if (!forceReplace)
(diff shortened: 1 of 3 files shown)


sirskunkalot:
Test build: https://github.com/Valheim-Modding/Jotunn/suites/2646342518/artifacts/58134315

JotunnLib/Managers/LocalizationManager.cs
```diff
@@ -211,9 +213,9 @@ public void AddToken(string token, string value, string language = DefaultLangua
         {
             Dictionary<string, string> languageDict = null;
 
-            if (token.Any(x => Localization.instance.m_endChars.Contains(x)))
+            if (token.Any(x => LocalizationEndChars.Contains(x)))
             {
-                throw new Exception($"Token has an end char in it ({Localization.instance.m_endChars}).");
+                throw new Exception($"Token has an end char in it ({LocalizationEndChars}).");
```
sirskunkalot: that error message :D of course it has. if it is not null it has an end char. but it might not be an "invalid" one.

Algorithman (approved):

PR #201, opened 2021-05-06
Docs: added example code and explanation for configuration entry read…

Algorithman:
…/write


RatikKapoor (approved):

sirskunkalot (approved):
needs to be changed in modexample, too

PR #210, opened 2021-05-10
BetterTrader Compatibility

Digitalroot:
Added a call to BetterTrader.ObjectDB_Patches.RegisterObjectDBItems if the BetterTrader mod is being used.

### Issue
Custom items added via Jotunn are loaded too late for BetterTrader to be able to use them. From what I can tell the way Jotunn injects into the loading process by using the DetourContext does not respect <a href="https://harmony.pardeike.net/articles/priorities.html">Harmony's Priority annotation</a>. Setting another mod to 
```
[HarmonyPriority(Priority.Last)]
```
or 
```
[HarmonyAfter(new string[]
{
	"com.jotunn.jotunn"
})]
```
Will still cause the mod to load before Jotunn.

### Conclusion
While this PR will allow BetterTrader to work with Items added via Jotunn, it does expose a compatibility problem with any mods that need to work with a complete set of items available to the game. 

e.g. The mod <a href="https://www.nexusmods.com/valheim/mods/1143">SteelAge</a> and <a href="https://www.nexusmods.com/valheim/mods/1179">RuneSwords</a> by zarboz (@sbtoonz) use Jotunn and are not available in BetterTrader without this PR. Looking at <a href="https://github.com/RandyKnapp/ValheimMods/tree/main/EpicLoot">EpicLoot</a> by @RandyKnapp. It also iterates over ObjectDB.instance.m_items. While I have not tested EpicLoot yet, I bet the new items added via SteelAge and RuneSwords will not drops as Magic/Rare...etc.


xiaoxiao921:
Thanks for reporting this to us, it seems like an upstream compatibility problem between HarmonyX and Monomod at first glance
I've made a report in the bepinex discord, we'll see where this go

Digitalroot:
> 
> 
> Thanks for reporting this to us, it seems like an upstream compatibility problem between HarmonyX and Monomod at first glance
> I've made a report in the bepinex discord, we'll see where this go

I agree the root cause is this and getting it fixed would be the best solution.

Windows10CE:
Hi there, one of the HarmonyX contributors here. I'll give a quick rundown of what's been said in the BepInEx Discord.

HarmonyX uses MonoMod to apply its patches, allowing both to coexist, but Harmony patches are still applied in a different manner than other MM detours.

Whenever you create a detour in MM (or use one of the MMHook events to do the same) your method will be added to the "detour chain" of the method you're hooking. When the method is called, the first detour in the chain is called, and when that method calls `orig` the next detour is called, and so on until you get to the original (or ILHooked) version of the method.

HarmonyX doesn't make patches part of this detour chain structure and instead opts to use one ILHook to apply all the patches for a method. This is the cause of the problem here, HarmonyX patches will always be run after the last `orig` call and always run before control is transferred back to any detour.

There is really only one option to resolve this on the HarmonyX end, that being to convert all patches to be their own separate detours in the MonoMod chain. This was considered back when HarmonyX was started but was later decided against as it would've required a lot of work for what seemed not worth the trouble at the time.

As for a solution to this specific problem in the present, my suggestion would either be to change this PR to be less specific to one mod (mod-specific code in a general library is a bit... off) by possibly making an event mods could subscribe to for after Jotunn has finished doing its thing.

Another option is to leave Jotunn out of this entirely, discard the PR,  and have BetterTrades postfix the Jotunn detour. It's admittedly a bit of a hack, but it also doesn't require any intervention or code on Jotunn's behalf.

sirskunkalot:
Thank you Win.

We do actually have events in Jotunn which can be subscribed by mods which need to run after Jotunn inserting Items etc. That would mean the mod must have a SoftDependency on Jotunn. It is not required for the mod to actually depend on Jotunn being there and distribute it alongside. It might be considered as taking ourself as more important as we are but I am totally with Win's opinion on having specific mod related workarounds which could be easily circumvented in that specific mod. Since there is no way to have MMHooks and HarmonyX patches respect each others priority wishes.

That is just my personal opinion and not what we discussed as the whole team. We have not discussed it until now.

Digitalroot:
> Another option is to leave Jotunn out of this entirely, discard the PR, and have BetterTrades postfix the Jotunn detour. It's admittedly a bit of a hack, but it also doesn't require any intervention or code on Jotunn's behalf.

FYI, BetterTraders source is not public.

Digitalroot:
I am happy with the current conversation. I agree 100% with the comments about “mod-specific code in a general library”. Jotunn should be mod agnostic and non-Jotunn mods should also be Jotunn unaware. If I was the developer of these mods, I would take this stance. 

However, as a user of Jotunn mods and non-Jotunn mods I do care that they work together or at least adhere to a consistent standard. I understand the root cause is that, “MMHooks and HarmonyX patches [do not] respect each other’s priority wishes.”

Until another solution becomes available, I am publishing https://github.com/Digitalroot/digitalroot-valheim-mods/blob/main/Digitalroot.Valheim.JotunnBetterTraderFix/Main.cs as a compromise/workaround for the specific issue I have raised in this PR. 

Thank you @sirskunkalot for pointing out the events in the ItemManager.

paddywaan:
We already expose [`ItemManager.OnItemsRegistered`](https://github.com/Valheim-Modding/Jotunn/blob/250cb650cdc43eb8b8f25c3670e9cdb079ff899c/JotunnLib/Managers/ItemManager.cs#L50) which is triggered after all mods have finished adding to the objectDB. @Digitalroot should consider subscribing to this event instead.

This issue is largely outside of our scope for a fix, and a specific fix for a specific mod isn't actually solving anything for anyone other than OP. Just use the event provided instead of introducing a bandaid mod-specific fix.

As far as the mod you have released goes.... You have:
```[BepInDependency(Jotunn.Main.ModGuid, BepInDependency.DependencyFlags.HardDependency)]``` a dependency on jotunn, so anyone who also depends on your mod, also depends on Jotunn. Why exactly you are introducing an intermediary dependency into the chain of dependencies instead of simply using the event is beyond me. You are not eliminating any dependency issues by doing so.

**Comment by sirskunkalot** [comment]
Hi digitalroot,

> Until another solution becomes available, I am publishing https://github.com/Digitalroot/digitalroot-valheim-mods/blob/main/Digitalroot.Valheim.JotunnBetterTraderFix/Main.cs as a compromise/workaround for the specific issue I have raised in this PR.
Thanks, that is a good workaround for now, i guess.

Did you try to reach out to the BetterTrader devs? It really is a shame when they don't publish their mods cause it always will be a community effort and that makes it fairly hard. 

I will close this PR for now. But if you have news on that matter please share them with us. I also will try to compress the information gathered here into a page for the Valheim Modding wiki when I find the time to do so today.

**Comment by Digitalroot** [comment]
> Did you try to reach out to the BetterTrader devs? It really is a shame when they don't publish their mods cause it always will be a community effort and that makes it fairly hard.

Yes, I have been in contact with the BetterTrader dev. They went as far as the following. (dnSpy)
```
	[HarmonyPriority(2147482648)]
	[HarmonyAfter(new string[]
	{
		"randyknapp.mods.epicloot",
		"com.jotunn.jotunn"
	})]
```
Which didn't resolve the issue and caused me to look deeper into it.

> I will close this PR for now. But if you have news on that matter please share them with us. I also will try to compress the information gathered here into a page for the Valheim Modding wiki when I find the time to do so today.

Thank you

PR #213, opened 2021-05-12
Added missing text from #180

paddywaan:
see #180


sirskunkalot (changes requested):
has conflicts which cant be resolved here on gh

paddywaan:
? How does this have conflicts, i pulled from dev.

Edit: Nevermind, apparently I did not, not entirely sure what has gone wrong there, i'll just resub a new pr.

PR #214, opened 2021-05-12
Fix #180

paddywaan:


Algorithman (commented):
Why remove the mention of our prebuild task?

paddywaan:
Add it back in if you want, this was just copied/pasted from #180, forget why but likely because I had reordered it elsewhere, or was unnecessary in the context. Can't say that I thoroughly looked over the document, but the mmhook detail is necessary to include regardless, as a user had pointed out that it is confusing without that detail in a quickstart.

sirskunkalot:
I've included some sections to make it even more clear.

sirskunkalot (approved):

Algorithman (approved):

PR #218, opened 2021-05-15
Make FixReferences not throw on Dictionary fields

xiaoxiao921:


JotunnLib/Managers/PieceManager.cs
```diff
@@ -257,7 +257,7 @@ private void RegisterInPieceTables()
                 }
                 catch (Exception ex)
                 {
-                    Logger.LogError($"Error while adding piece {customPiece}: {ex.Message}");
+                    Logger.LogError($"Error while adding piece {customPiece}: {ex}");
```
sirskunkalot, line 260: can you include a change to the itemmanager (full ex) in this PR, too?

sirskunkalot (changes requested):

sirskunkalot (approved):

Algorithman (approved):

PR #219, opened 2021-05-16
Added SimpleJVL.cs  to Jotunn.Utils

Mr-Rageous:


JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,211 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using UnityEngine;
+using Jotunn.Managers;
+using Jotunn.Entities;
+using Jotunn.Configs;
+
+namespace Jotunn.Utils
+{
+    static class SimpleJVL
```
xiaoxiao921: always specify the access modifier

xiaoxiao921 (commented):
I'm personally not a big fan on having all those functions put into the same class instead of being seperated but if the majority think its better this way then all good

xiaoxiao921:
Code should also have a doc page related to it or atleast xml docs

JotunnLib/Utils/SimpleJVL.cs
Mr-Rageous: Is there a preference here whether I make it public or private?

xiaoxiao921: must be public, like the methods, otherwise people that uses Jotunn won't be able to access the SimpleJVL class and the methods inside it

Mr-Rageous:
> @xiaoxiao921:  Code should also have a doc page related to it or at least xml docs

I began it, I will continue through-out today. Do I need to write myself into README and such or would one of you do that?

xiaoxiao921:
> > @xiaoxiao921:  Code should also have a doc page related to it or at least xml docs
> 
> I began it, I will continue through-out today. Do I need to write myself into README and such or would one of you do that?

Add yourself please

Also please we require to have every field / method to have the access modifier specified, by default in C# the default access modifier when not written is internal, meaning that the methods currently cannot be used by users because they are marked as internal

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,217 @@
...
+        public static CraftingStation reflect;
+        public static string needs;
+        public static int needsAmount;
+        public static int needsAmountPerLevel;
+        public static bool needsRecovery;
+        public static int amount;
+        public static int amountPerLevel;
+        public static void LoadAssets(string assetBundleName)
+        {
+            AssetBundle.UnloadAllAssetBundles(false);
```
xiaoxiao921: This will unload all asset bundles, potentially from other mods too, why is it there ?

Mr-Rageous: To prevent the error from trying to load multiple assetbundles while previous bundles are still loaded, how would I fix this to simply unload the current bundle if a bundle is already loaded?

JotunnLib.sln
```diff
@@ -27,6 +27,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "resources", "resources", "{
 		README.md = README.md
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{92BDA092-A795-41F0-803F-A683BBF3BD09}"
```
sirskunkalot, line 30: Please dont add items to the solution

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,277 @@
...
+        public static string needs;
+        public static int needsAmount;
+        public static int needsAmountPerLevel;
+        public static bool needsRecovery;
+        public static int amount;
+        public static int amountPerLevel;
+        public static void LoadAssets(string assetBundleName)
+        {
+            if (assetBundle) assetBundle.Unload(false);
+            Jotunn.Logger.LogInfo($"Embedded resources: {string.Join(",", typeof(Type).Assembly.GetManifestResourceNames())}");
```
sirskunkalot: That can't work, typeof(Type) yields the Type type. I am not entirely sure whats the best approach here since the resources will always be embedded in the resulting dll, not in Jötunn

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,277 @@
...
+namespace Jotunn.Utils
+{
+    public static class SimpleJVL
+    {   // Simple JVL - For simpletons, a language simplifier
+        public static AssetBundle assetBundle;
+        public static GameObject prefabObject;
+        public static ItemDrop.ItemData itemDrop;
+        public static CraftingStation reflect;
+        public static string needs;
+        public static int needsAmount;
```
sirskunkalot: I still dont know why there are members here. They will be null if not set from the outside before.

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,277 @@
...
+                new PieceConfig
+                {
+                    PieceTable = pieceTable,
+                    CraftingStation = craftingStation,
+                    AllowedInDungeons = false,
+                    Requirements = new[]
+                    {
+                        new RequirementConfig
+                        {
+                            Item = needs,
```
sirskunkalot: Will be null if class member is not set

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,277 @@
...
+                            Amount = amount,
+                            AmountPerLevel = amountPerLevel,
+                            Recover = needsRecovery
+                        }
+                    }
+                });
+            PieceManager.Instance.AddPiece(piece);
+        }
+        public static void AddOneSlotRecipe(GameObject prefabNew, bool fixRefs, string craftingStation, string repairStation, int minStationLevel, int amount, string needs1, int needsAmount1, int needsAmountPerLevel1)
```
sirskunkalot: That Recipe methods are just overhead imho. Use an Array of string if you want to, but having four methods with three parameters per requirement is not good.

JotunnLib/Documentation/home/tutorials/pieces.md
```diff
@@ -80,3 +80,18 @@ And here we have our final results:<br />
 ![Blue Print Rune Piece Table](../../images/data/BluePrintRunePieceTable.png)
 
 As you can see in the screenshot the name and description are not yet localized. This is done via the `blueprintRuneLocalizations()` method. To read more about localization/translation head over to the [localization tutorial pages](localization.md).
+
+## Using SimpleJVL
```
sirskunkalot: Please dont add them to the "normal" tutorials but just to the "global" SimpleJVL docs.

JotunnLib/Documentation/home/tutorials/item-conversions.md
```diff
@@ -80,4 +80,23 @@ var blastConversion = new CustomItemConversion(new SmelterConversionConfig
     ToItem = "Steel" // This is our custom prefabs name
 });
 ItemManager.Instance.AddItemConversion(blastConversion);
+```
+
+## Using SimpleJVL
```
sirskunkalot: Please dont add them to the "normal" tutorials but just to the "global" SimpleJVL docs.

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,277 @@
...
+                new ItemConfig
+                {
+                    Amount = amount,
+                    CraftingStation = craftingStation,
+                    MinStationLevel = minStationLevel,
+                    Requirements = inputs
+                });
+            ItemManager.Instance.AddItem(recipe);
+        }
+        public static void AddPieceRecipe(GameObject pieceName, string pieceTable, string craftingStation, params RequirementConfig[] inputs)
```
sirskunkalot: There is no such thing as a Piece Recipe. Pieces just have requirements while Recipes are not only used for crafting but for upgrading of items, too.

sirskunkalot (changes requested):
Also there is a README.bbcode which needs to be updated.

Honestly that code is not in a good shape and as discussed on Discord before there are many obstacles not even touched (like if fixReferences for prefabs must be true or false).

I like some of the abstractions for ease of use, but in the end it is a complex field and you cannot have everything abstracted.

JotunnLib/Utils/SimpleJVL.cs
baruchvlz: Isn't this the same functionality that in is [Jotunn.Utils.AssetUtils](https://github.com/Valheim-Modding/Jotunn/blob/dev/JotunnLib/Utils/AssetUtils.cs#L104)?

JotunnLib/Utils/SimpleJVL.cs
Mr-Rageous: Now named PieceRequirement, are you also saying this is an unnecessary method that should be a part of the method it is meant to be used with generally?

JotunnLib.sln
Mr-Rageous: Probably a mistake, my bad

Mr-Rageous:
Should I convert this into a Draft for now?

Mr-Rageous:
> Also there is a README.bbcode which needs to be updated.
> 
> Honestly that code is not in a good shape and as discussed on Discord before there are many obstacles not even touched (like if fixReferences for prefabs must be true or false).
> 
> I like some of the abstractions for ease of use, but in the end it is a complex field and you cannot have everything abstracted.

Hopefully it looks much better now.

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -258,36 +258,161 @@ public void AddClonedItem(string prefabNew, string prefabOld, string name, strin
...
+        /// <param name="foodRegen">How much Health this item grants per second.</param>
+        /// <param name="foodStamina">How much Stamina this item grants per second.</param>
+        /// <param name="setName">Name of the set that this item belongs to.</param>
+        /// <param name="setSize">Size of the set that this item belongs to.</param>
+        /// <param name="setStatusEffect">Triggered <see cref="StatusEffect"/> when player has equipped <paramref name="setSize"/> of <paramref name="setName"/></param>
+        /// <param name="durabilityDrain">Durability reduced for each use.</param>
+        /// <param name="durabilityPerLevel">Additional <paramref name="durabilityDrain"/> per upgrade of this item.</param>
+        /// <param name="useDurability">Durability reduced for each use.</param>
+        /// <param name="useDurabilityDrain">Whether or not this item uses <paramref name="durabilityDrain"/> when actions are performed.</param>
+        public void AddClonedItem(string prefabNew, string prefabOld, string name, string description,
```
sirskunkalot: The method of Mona Lisa ;)

Mr-Rageous: Hey, I considered it thoroughly before implementing it that way. If the constructor has 1 overload that points to this, the user doesn't get the spam from IntelliSense, but functionality stays the same, additionally allows the author to target specific fields while leaving others untouched simply by adding on, which creates the overload situation into the optional params. I couldn't find a better way to get this kind of wrapping, but I am open to suggestions.

Mr-Rageous: Removed _The Methods of Mona Lisa_ until further notice.

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+        ///     Adds a <see cref="Piece"/> to ObjectDB with inherited properties by default, and builds a new recipe for it.
+        /// </summary>
+        /// <param name="prefabName">Case sensitive name of the new <see cref="Piece"/>.</param>
+        /// <param name="name">Token name of the new item.</param>
+        /// <param name="description">Description of the item, shows up in tooltip.</param>
+        /// <param name="pieceTable">Case sensitive name of <see cref="PieceTable"/> assigned to the recipe.</param>
+        /// <param name="craftingStation">Case sensitive name of the <see cref="CraftingStation"/> required </param>
+        /// <param name="inputs">Recipe config as a <see cref="RequirementConfig"/></param>
+        public void AddPiece(string prefabName, string name, string description, GameObject prefab,
+            string pieceTable = "piece_HammerPieceTable", string craftingStation = "",
```
sirskunkalot: its just "_HammerPieceTable"

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+            ItemManager.Instance.AddItem(Item);
+        }
+
+        /// <summary>
+        ///     Adds a <see cref="Piece"/> to ObjectDB with inherited properties by default, and builds a new recipe for it.
+        /// </summary>
+        /// <param name="prefabName">Case sensitive name of the new <see cref="Piece"/>.</param>
+        /// <param name="name">Token name of the new item.</param>
+        /// <param name="description">Description of the item, shows up in tooltip.</param>
+        /// <param name="pieceTable">Case sensitive name of <see cref="PieceTable"/> assigned to the recipe.</param>
```
sirskunkalot, line 219: No recipes with Pieces

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// <param name="craftingStation">Case sensitive name of the <see cref="CraftingStation"/> required </param>
+        /// <param name="inputs">Recipe config as a <see cref="RequirementConfig"/></param>
+        public void AddPiece(string prefabName, string name, string description, GameObject prefab,
+            string pieceTable = "piece_HammerPieceTable", string craftingStation = "",
+            bool isAllowedInDungeons = false, params RequirementConfig[] inputs)
+        {
+            AddStationPiece(prefabName, name, description);
```
sirskunkalot: why does it always make a station out of it?

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+        public void AddStationPiece(string prefabName, string name, string description)
+        {
+            GameObject prefabObject = AddItem(prefabName, name, description);
+            CraftingStation reflect = prefabObject.GetComponent<CraftingStation>();
+            if (reflect == null)
+            {
+                GameObject er = GameObject.Find("piece_workbench");
+                CraftingStation craft = er.GetComponent<CraftingStation>();
+                prefabObject.AddComponent<CraftingStation>();
+                CraftingStation Craft = prefabObject.GetComponent<CraftingStation>();
```
sirskunkalot: Why not just clone the CraftingStation component instead of copying field by field? Hard to maintain...

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+            {
+                case "piece_cookingstation":
+                    conversion = new CustomItemConversion(new CookingConversionConfig {Station = station, FromItem = fromitem, ToItem = toitem});
+                    ItemManager.Instance.AddItemConversion(conversion);
+                    break;
+                case "fermenter":
+                    conversion = new CustomItemConversion(new FermenterConversionConfig {Station = station, FromItem = fromitem, ToItem = toitem});
+                    ItemManager.Instance.AddItemConversion(conversion);
+                    break;
+                case "smelter":
```
sirskunkalot, line 331: Lacks 60% of the conversions, does not work with custom conversion pieces (magic strings)

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// </summary>
+        /// <param name="pieceName">Case sensitive name of the new .prefab.</param>
+        /// <param name="pieceTable">Case sensitive name of the new .prefab.</param>
+        /// <param name="craftingStation">Case sensitive name of the <see cref="CraftingStation"/> required for crafting this item.</param>
+        /// <param name="isAllowedInDungeons">Amount of this item that <see cref="RequirementConfig"/> will return.</param>
+        /// <param name="inputs">Recipe config as a <see cref="RequirementConfig"/></param>
+        public void AddPieceRequirements(string pieceName, string pieceTable, string craftingStation,
+            bool isAllowedInDungeons = false, params RequirementConfig[] inputs)
+        {
+            CustomPiece piece = new CustomPiece(pieceName,
```
sirskunkalot: This creates an "empty" piece which is just a cube model with nothing attached

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        /// <summary>
+        ///     Adds a <see cref="RecipeConfig"/> for a <see cref="Piece"/>.
+        /// </summary>
+        /// <param name="pieceName">Case sensitive name of the new .prefab.</param>
+        /// <param name="pieceTable">Case sensitive name of the new .prefab.</param>
+        /// <param name="craftingStation">Case sensitive name of the <see cref="CraftingStation"/> required for crafting this item.</param>
+        /// <param name="isAllowedInDungeons">Amount of this item that <see cref="RequirementConfig"/> will return.</param>
+        /// <param name="inputs">Recipe config as a <see cref="RequirementConfig"/></param>
+        public void AddPieceRequirements(string pieceName, string pieceTable, string craftingStation,
```
sirskunkalot: Name is misleading

JotunnLib/Utils/SimpleJVL.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// <param name="minStationLevel">Level of the <see cref="CraftingStation"/> required for crafting this item.</param>
+        /// <param name="amount">Amount of this item that <see cref="RequirementConfig"/> will return.</param>
+        /// <param name="inputs">Recipe config as a <see cref="RequirementConfig"/></param>
+        public void ChangeRecipe(string prefabOld, bool fixRefs = false, string craftingStation = "",
+            string repairStation = "", int minStationLevel = 1, int amount = 1,
+            params RequirementConfig[] inputs)
+        {
+            GameObject ger = GameObject.Find(prefabOld);
+            ItemDrop change = ger.GetComponent<ItemDrop>();
+            ItemData d = change.GetComponent<Recipe>();
```
sirskunkalot: Is typed ItemData but gets the Recipe component

sirskunkalot (changes requested):
Probably missed some but that's what I could see just from reading.

That is nowhere usable. Please test new PRs at least once before letting us review it. This would not even compile.

PR #280, opened 2021-08-24
Update PieceConfig.cs

Dominowood371:
The description for pieces was not displayed using Properties Description


sirskunkalot (approved):
Thanks, nice find. Seems nobody used that before ;)

PR #282, opened 2021-08-27
load text file from resources

MSchmoecker:
Adds a method to read text from an embedded resource. For example it is useful for language files. Additionally it adds a hint to the point separation of an embedded resources.


sirskunkalot:
thx

PR #283, opened 2021-08-28
More ApplyStyle for UI

MSchmoecker:
This exposes the woodpanel styling as a public method and adds styling for ScrollRect and Scrollbar components. Sometimes you may only need Scrollbar so both methods are public.

grafik

### DIFF
--- a/JotunnLib/Managers/GUIManager.cs
+++ b/JotunnLib/Managers/GUIManager.cs
@@ -1447,5 +1447,67 @@ namespace Jotunn.Managers
                 checkbox.maskable = true;
             }
         }
+
+        /// <summary>
+        ///     Apply Valheim style to a <see cref="ScrollRect"/> component.
+        /// </summary>
+        /// <param name="scrollRect"> Component to apply the style to </param>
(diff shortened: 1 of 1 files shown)


sirskunkalot:
It failed to compile. I guess you used language features not supported by the dotnet version I use in the github workflow... I tried to update to a new version (even beta 6.0) but I could not get it to work. If you have any idea what to do (I think you use Linux, right?), I would be glad...

MSchmoecker:
Hopefully it compiles now :)

sirskunkalot:
Looks good. Still wondering how to achieve this ;) It does not like "SomeClass bla = new()", either...

MSchmoecker:
I haven't worked with Github action jet so I have no idea

sirskunkalot:
It's using ubuntu with this toolset:
sudo apt-get install -y mono-roslyn mono-complete mono-dbg msbuild unzip dirmngr dotnet-sdk-5.0 dotnet-runtime-5.0
And that doesn't like some new features.

Anyway, I will check out your PR soon.

sirskunkalot:
looks good, thx

PR #289, opened 2021-09-15
added ApplyTextStyle

MSchmoecker:
Adds 
* `ApplyTextStyle(Text text, Font font, Color color, int fontSize = 16, bool createOutline = true)`
* `ApplyTextStyle(Text text, Color color, int fontSize = 16, bool createOutline = true)`
* `ApplyTextStyle(Text text, int fontSize = 16)`

By default, is always font=AveriaSerifBold, color=Color.white and createOutline=true. These get used inside GUIManager and are not changing any existing fontSize


sirskunkalot:
thx!

PR #294, opened 2021-09-17
Create bug_report

paddywaan:


sirskunkalot:
Good idea. Maybe add a "mod List / used mod" section. And also "Link to LogOutput.log dump"?

PR #301, opened 2021-09-29
Added slider to InGameConfig

MSchmoecker:
Displays a slider to int, float and double fields if a `AcceptableValueRange<type>` is present. The slider is spawned for every InputField but only activated if needed. Also adds a basic `ApplySliderStyle` to GUIManager.

grafik


sirskunkalot:
Very nice. Left of the text input might render the list very long with long descriptions, I guess (VNEI for example :P). Have you tried putting the slider below the text input?

MSchmoecker:
This could be a problem, right. It would overlap if the key is roughly over 45 chars long. Below the text field is always enough space, I will put it there

sirskunkalot:
grafik
might truncate that to three digits, i guess?! i can quickly implement that, if you want. or do you think there is a use case for seven digit precision?

PR #311, opened 2021-10-23
ZoneManager

MathiasDecrock:
fails on 2nd login of world in same game session


MathiasDecrock:
A large amount of changed files, but most are new images & tiny cosmetic changes to autogenerated documentation markdown files (use the filters in the changes tab to exclude pngs)

Check TestMod for usage

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -1,47 +1,300 @@
...
+                FixMockReferences(transform);
+            }
             
-            Logger.LogInfo("Registering custom zone data");
+            return locationContainer;
+        }
 
-            // Call event handlers to load prefabs
-            OnBeforeZoneLoad?.Invoke(null, EventArgs.Empty);
+        private void FixMockReferences(Transform transform)
```
sirskunkalot: Why did you write your own mocking code? GameObject.FixReferences() would have been fine, also traversing the child GOs

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -1,47 +1,300 @@
...
+        ///     Create a CustomLocation that is a deep copy of the original.
+        ///     Changes will not affect the original. The CustomLocation is already registered to be added.
+        /// </summary>
+        /// <param name="name">name of the custom location</param>
+        /// <param name="baseName">name of the existing location to copy</param>
+        /// <returns></returns>
+        public CustomLocation CreateClonedLocation(string name, string baseName)
+        {
+            var baseZoneLocation = GetZoneLocation(baseName);
+            var copiedPrefab = Object.Instantiate(baseZoneLocation.m_prefab, Vector3.zero, Quaternion.identity, LocationContainer.transform);
```
sirskunkalot, line 204: Should use PrefabManager, i guess

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -1,47 +1,300 @@
...
+        {
+            if (Locations.TryGetValue(customLocation.Name, out CustomLocation existingLocation))
             {
-                ZoneSystem.instance.m_vegetation.Add(veg);
-                Logger.LogDebug("Added vegetation: " + veg.m_name);
+                Logger.LogWarning($"Location {customLocation.Name} already exists");
+                return false;
             }
+
+            customLocation.Prefab.transform.SetParent(LocationContainer.transform);
```
sirskunkalot: Please add the Prefab to the PrefabManager here. All Managers behave like that.

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -1,47 +1,300 @@
...
+        /// <summary>
+        ///     Register a CustomVegetation to be added to the ZoneSystem
+        /// </summary>
+        /// <param name="customVegetation"></param>
+        /// <returns></returns>
+        public bool AddCustomVegetation(CustomVegetation customVegetation)
         {
-            Vegetation.Add(veg);
+            Logger.LogDebug($"Registering custom vegetation {customVegetation.Name}");
+            Vegetations.Add(customVegetation.Name, customVegetation);
```
sirskunkalot, line 249: Add to PrefabManager here, too

JotunnLib/Managers/SynchronizationManager.cs
```diff
@@ -442,6 +442,9 @@ private void FejdStartup_Awake(On.FejdStartup.orig_Awake orig, FejdStartup self)
             // Harmony patch BepInEx to ensure locked values are not overwritten
             Harmony.CreateAndPatchAll(typeof(ConfigEntryBase_SetSerializedValue));
             Harmony.CreateAndPatchAll(typeof(ConfigEntryBase_GetSerializedValue));
+#if DEBUG
+            Harmony.CreateAndPatchAll(typeof(Debug_isDebugBuild));
```
sirskunkalot: Why is this in the SyncManager? Please add to DebugUtils. That class is already guarded by the #Debug pragma

JotunnLib/Managers/MockManager.cs
```diff
@@ -82,16 +82,13 @@ private void RemoveMockPrefabs(On.ObjectDB.orig_Awake orig, ObjectDB self)
         {
             orig(self);
 
-            if (SceneManager.GetActiveScene().name == "main")
+            if (SceneManager.GetActiveScene().name == "main" && MockPrefabContainer.transform.childCount > 0)
```
sirskunkalot: Is not necessary, foreach() does not enumerate where there is none. Or is there any special intent?

sirskunkalot (changes requested):
Small review of what I found at first glance

JotunnLib/Managers/ZoneManager.cs
MathiasDecrock: I couldn't find where complete GameObjects were replaced? I would need the same "copyRegex" there, to allow multiple copies of the same mock (or you'd need to do a lot of nesting workarounds)

JotunnLib/Managers/ZoneManager.cs
MathiasDecrock: Locations aren't retrieved from ZNetScene, but are in the ZoneSystem. Vegetation needs to be in the PrefabManager though. Probably not an issue to add it though, pretty much the same. Another opportunity for name collisions though.

JotunnLib/Managers/MockManager.cs
MathiasDecrock: I have no idea when or why I did this, I'll remove it :P

JotunnLib/Managers/ZoneManager.cs
MathiasDecrock: Moved to FixReference extension method

JotunnLib/DebugUtils/DebugHelper.cs
```diff
@@ -44,6 +45,9 @@ private void Awake()
                 catch (Exception) {}
             };
             On.ZNet.RPC_ClientHandshake += ProvidePasswordPatch;
+#if DEBUG
```
sirskunkalot: redundant, the whole DebugHelper.cs is gated by #DEBUG

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -259,7 +259,14 @@ public bool AddCustomLocation(CustomLocation customLocation)
                 return false;
             }
 
-            customLocation.Prefab.transform.SetParent(LocationContainer.transform);
+            customLocation.Prefab.transform.SetParent(LocationContainer.transform); 
+            foreach(var zNetView in customLocation.ZoneLocation.m_netViews)
+            { 
+                if (!PrefabManager.Instance.Prefabs.ContainsKey(zNetView.gameObject.name.GetStableHashCode()))
```
sirskunkalot, line 227: why is it done manually and not via PrefabManager.AddPrefab? that would give out a warning when the name is redundant. or is it common that prefabs are redundant in a zone?

sirskunkalot (changes requested):
some small stuff

JotunnLib/Managers/ZoneManager.cs
MathiasDecrock: Yeah, it's likely a normal prefab that is already registered, only in rare cases would this actually be needed, if you add a custom prefab only in a location, but don't register it yourself (like an altar you can't build or something)
Most prefabs in the location will be normal things like walls & floors, which are already registered

sirskunkalot (approved):

PR #317, opened 2021-10-25
feat: Add IsCheat to ConsoleCommand

### LINKED ISSUE #315: [FEATURE] Allow custom commands to set the isCheat flag
**User Story**
As a mod developer, I want to add a custom command that is hidden behind "devcommands".

**Envisioned Solution**
This should be easily done by adding a field to Jotunn's `ConsoleCommand` that corresponds to the `Terminal.ConsoleCommand.IsCheat` flag.

**Additional context**
None that I can think of.

I will put together a pull request if there are no objections.

joeyparrish:
Closes #315

### DIFF
--- a/JotunnLib/Entities/ConsoleCommand.cs
+++ b/JotunnLib/Entities/ConsoleCommand.cs
@@ -16,6 +16,26 @@
         public abstract string Help { get; }
 
         /// <summary>
+        ///     If true, this command will only work after `devcommands` is run in the console.
+        /// </summary>
+        public virtual bool IsCheat => false;
+
+        /// <summary>
+        ///     If true, this command will be allowed in networked play.
+        /// </summary>
+        public virtual bool IsNetwork => false;
+
+        /// <summary>
+        ///     If true, and IsNetwork is true, this command will be allowed in networked play, but only for the server.
+        /// </summary>
+        public virtual bool OnlyServer => false;
+
+        /// <summary>
+        ///     If true, this command will not be shown when the user types `help` into their console.
+        /// </summary>
+        public virtual bool IsSecret => false;
+
+        /// <summary>
         ///     The function that will be called when the user runs your console command, with space-delimited arguments.
         /// </summary>
         /// <param name="args">The arguments the user types, with spaces being the delimiter.</param>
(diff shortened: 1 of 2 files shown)


joeyparrish:
I'll try to verify it locally using the toolchain setup from GitHub Actions.  (Very helpful tip, BTW.)  In the meantime, here it is.  Thanks!

JotunnLib/Entities/ConsoleCommand.cs
```diff
@@ -15,6 +15,11 @@ public abstract class ConsoleCommand : CustomEntity
         /// </summary>
         public abstract string Help { get; }
 
+        /// <summary>
+        ///     If true, this command will only work after `devcommands` is run in the console.
+        /// </summary>
+        public abstract bool IsCheat { get; }
```
sirskunkalot: Introducing new abstract properties renders current inheriting classes invalid (and thus all mods using console commands). I'd rather not introduce breaking changes here. This should probably just be a concrete base class property with a reasonable default (vanilla defaults should be fine).

sirskunkalot: Another approach would be to have the CommandManager's AddConsoleCommand take optional parameters to change the commands flags.

joeyparrish: I finally have the tool chain working locally.  Sorry for uploading a broken PR initially!

I just tried making IsCheat non-abstract, but that doesn't quite work, either.  (I'm new to C#.)

I'm going to take your advice to add a parameter to AddConsoleCommand.  I'll upload new code shortly once I've fully tested it.

Thanks!

joeyparrish: Looks like I can make IsCheat virtual, rather than abstract, and give it a default of false, preserving both backward compatibility and the style of the other fields.

What do you think?

sirskunkalot: Yes, sounds good. If you are on it anyway, could you also implement the other vanilla properties (isSecret, etc)?

joeyparrish: Sure.  I wasn't familiar with those, but I'll check it out.

joeyparrish: Done!  Please take a look.

sirskunkalot (approved):
looks good, thank you.

sirskunkalot:
released with v2.3.9

PR #321, opened 2021-10-28
fix: Fix Dropdown width setting

### LINKED ISSUE #320: [BUG] Dropdown width only affects inner contents
Details:
Jotunn Version: 2.3.9
Jotunn Submodule(if applicable): GUIManager
Repeatability(Consistent(100%), Inconsistent(50%), Rare(1%), Unknown(==1)): 100%

**Problem Description:**
Setting the width of a Dropdown only affects its contents, not the dropdown itself.

**Expected Behaviour:**
I expect the Dropdown and its inner contents to be sized appropriately, with the explicit width I set to be the width of the outer element.

**Actual Behaviour:**
The outer element is always exactly 160px wide.  The inner contents are resized, causing weird effects.  The selection indicator (little white ball) is always in the same location.  When the inner contents are wider than 100px, they overlap with this indicator. 
 See screenshot:

dropdown-bug

I believe I have a fix for this, so I will send a PR soon.

joeyparrish:
Dropdown width now affects the outer dropdown component, and the inner
components are sized a little smaller to accommodate the selection
indicator.

Fixes #320

### DIFF
--- a/JotunnLib/Managers/GUIManager.cs
+++ b/JotunnLib/Managers/GUIManager.cs
@@ -180,6 +180,12 @@ namespace Jotunn.Managers
         private static int InputBlockRequests;
 
         /// <summary>
+        ///     The difference in width between a Dropdown and its contents, to allow
+        ///     room for the selection indicator.
+        /// </summary>
+        private const float DropdownContentOffset = 50f;
+
+        /// <summary>
         ///     Block all input except GUI
         /// </summary>
         /// <param name="state">Indicator if the input should be blocked or released</param>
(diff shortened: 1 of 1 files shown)


joeyparrish:
Here's what the fix looks like:

dropdown-fix

That's width: 180f.

JotunnLib/Managers/GUIManager.cs
```diff
@@ -179,6 +179,12 @@ public static bool IsHeadless()
         /// </summary>
         private static int InputBlockRequests;
 
+        /// <summary>
+        ///     The difference in width between a Dropdown and its contents, to allow
+        ///     room for the selection indicator.
+        /// </summary>
+        private const float DropdownContentOffset = 50f;
```
sirskunkalot, line 186: Doesn't need to be a const field, its single use and might just be in that method then.

sirskunkalot: Good catch otherwise ;) thanks

joeyparrish: Acknowledged.  Sorry for not responding to that before you merged.  We must be a few timezones apart.  :-)  Of course, you're always welcome to make changes as you see fit.

At the time, I thought it better to explain what this magic number was, so a documented constant felt right.  A local comment in the method would have done as well, though, I suppose.

joeyparrish:
Thank you!

JotunnLib/Managers/GUIManager.cs
sirskunkalot: No problem, It's not a big change, so I figured I will just do it myself quick instead of having you delete one line and push it again ;) Thanks for the fix, though. I remember seeing that it does behave weird but totally forgot about it.

PR #322, opened 2021-10-28
expose RenderRequest and add rotation

MathiasDecrock:


JotunnLib/Managers/RenderManager.cs
```diff
@@ -46,28 +55,39 @@ public void Init()
...
-        ///     Queues a new prefab to be rendered. The resulting <see cref="Sprite"/> will be ready at the next frame.
-        ///     If there is no active visual Mesh attached to the target, this method invokes the callback with null immediately.
+        ///     Create a <see cref="Sprite"/> of the <paramref name="target"/>
         /// </summary>
-        /// <param name="target">Object to be rendered. A copy of the provided GameObject will be created for rendering</param>
-        /// <param name="callback">Action that gets called when the rendering is complete</param>
-        /// <param name="width">Width of the resulting <see cref="Sprite"/></param>
-        /// <param name="height">Height of the resulting <see cref="Sprite"/></param>
-        /// <returns>Only true if the target was queued for rendering</returns>
-        public bool EnqueueRender(GameObject target, Action<Sprite> callback, int width = 128, int height = 128)
```
sirskunkalot, line 57: Thats public API, I would keep that method as an overload, [Obsolete] if needed

PR #325, opened 2021-10-30
Improved controller navigation in mod settings

koosemose:
Makes individual settings in Mod Settings navigable with a controller. Mostly addresses concerns I brought up in Issue Valheim-Modding/ModSettings#1 

Interacts with some settings better than others:

- Toggles work great.
- Inputs with sliders work reasonably well (depending on range, steps may be excessively large)
- Pure text inputs are intentionally skipped (avoids getting user trapped in text input)
- Color settings bring up the pop up, but can't be changed.
- Keybinds are a mixed result, slight changes were made to keep the keybind dialog from taking the button press that activates it as the keybind, but inputs are still passed to the settings (so confirm button is set as keybind, and immediately brings up a new keybind dialog, cancel button backs out of settings). Other buttons work fine.
- Gamepad button settings bring up the dropdown but the drop down's can't be navigated.

Also binds the navigation from Ok button and back Button to each other. I attempted to get navigation from them back to the settings, but failed.

While this is far from perfect, it addresses the primary goal of making the mod settings navigable by controller, is overall an improvement , and further improvement is beyond my current familiarity with the code base.

### DIFF
--- a/JotunnLib/GUI/InGameConfig.cs
+++ b/JotunnLib/GUI/InGameConfig.cs
@@ -264,6 +264,23 @@ namespace Jotunn.GUI
                     yield return CreateContent(mod, Configs[mod.Key]);
                 }
             }
+
+            // Connect OK and Back's Navigation
+            var okButton = ok.GetComponent<Button>();
+            var backButton = back.GetComponent<Button>();
+
+            var okNavigation = okButton.navigation;
+            var backNavigation = backButton.navigation;
+
+            okNavigation.selectOnLeft = backButton;
+            backNavigation.selectOnRight = okButton;
+
+            // These don't seem to work, leaving in to show intention
+            okNavigation.selectOnUp = Configs.Values.Last().gameObject.GetComponent<Selectable>();
+            backNavigation.selectOnUp = Configs.Values.Last().gameObject.GetComponent<Selectable>();
+
+            okButton.navigation = okNavigation;
+            backButton.navigation = backNavigation;
         }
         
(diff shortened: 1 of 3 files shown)


sirskunkalot (changes requested):
Thank you. Two things would be nice before merging it. Please tell me if you think you can do them or if I should try my luck. 

- Automatic scrolling of the scrollrect
- Some kind of visual indicator which setting is selected right now

koosemose:
The second would be trivial, most of the code is already there. For now I just chained the TargetGraphic of SelectableConfig to the TargetGraphic of the InteractionTarget and copy it's settings, so it does whatever those would do when selected (this is why the toggles light up fairly noticeably, whereas other things don't, they have more noticable settings already). However, we would need either an image that would do well with a color tint, sprites to swap between, or an animation. If there's something preexisting in valheim that would work I could hook it up easily, but if visual assets need to be created, I can't help with that.

On the first... I could maybe do it, but the dynamic nature of the list throws me off badly, and I would need a few days away from the code, then a few more days studying the code (and what the structure looks like in game, gameobject hierarchy-wise), so would prefer to either leave that for another PR, or have you try your luck.

sirskunkalot:
So, I went and recreated the mod settings in Unity. Unfortunately that rendered this PR obsolete. Good news is, that it might be more easy to support controller now, since the whole GameObject is self-made and can be customized without having vanilla stuff interfering. If you find the time to take a look at it, I would be glad. Otherwise I will put that on my to-do list (but maybe not at the highest priority).

PR #328, opened 2021-11-05
RenderManager render Sprite immediately

MSchmoecker:
RenderManager now is using `Object.DestroyImmediate` instead of `Object.Destroy` which is executed directly instead of taking one frame. This completely makes the callback irrelevant and the sprite can be used directly. Therefore the usage is a lot simpler as it can be seen in the TestMod.

For compatibility the new method is named `Render` and the old ones are marked as obsolete. The coroutine is still used to clear the renderer when not needed. As before, the same rendering Camera and Light is used when executing multiple renders in one frame

### DIFF
--- a/JotunnLib/Documentation/tutorials/renderqueue.md
+++ b/JotunnLib/Documentation/tutorials/renderqueue.md
@@ -18,8 +18,7 @@ private void Awake()
 }
 ```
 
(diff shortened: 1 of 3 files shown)


sirskunkalot (approved):
makes sense, thanks

PR #335, opened 2021-11-20
RenderManager copy components instead of destroying spawned components

MSchmoecker:
The previous solution was to spawn the whole prefab and then remove not needed components. Because components can depend on one another, this has lead to conflicts with RRR for example. Now the Prefab is recursively traversed and only Transform and visuals are copied to the spawned prefab.

~Because `SkinnedMeshRenderer` don't work well as they are copied (they depend on bone references and animators) they are faked with a `MeshRender`. This works most the time well, but can be off when the original mesh was rotated somehow unexpected. Though I only noticed this on Bonemass, so this might be fine?~

### DIFF
--- a/JotunnLib/Extensions/GameObjectExtension.cs
+++ b/JotunnLib/Extensions/GameObjectExtension.cs
@@ -1,4 +1,5 @@
-﻿using UnityEngine;
+﻿using System.Reflection;
+using UnityEngine;
 using UnityEngine.UI;
 
 namespace Jotunn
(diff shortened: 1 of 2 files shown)


sirskunkalot:
Would a BoneReorder on that Prefab help?

sirskunkalot:
<https://github.com/Valheim-Modding/Jotunn/blob/dev/JotunnLib/Utils/BoneReorder.cs>

Maybe that would need to expose some one-off methods for single prefabs then.

PR #338, opened 2021-12-04
fix: Fix CustomLocalization.Contains

joeyparrish:
Before, this was checking for a value instead of a key.

This also fixes the docs for the return.


sirskunkalot:
Thanks, good catch.

btw: You want to be listed as a contributor in the README? You contributed a lot by now (and its really appreciated).

joeyparrish:
Sure, sounds good to me.  I'll put that in my next PR.  :-)

joeyparrish:
What do the numbers mean next to people's names in the README?  For example, Tekla#1012 or blaxxun#9098.

sirskunkalot:
Oh, that is the discord username. They put numbers behind the name for uniqueness, i guess. So if you have a discord account, you could link it there, too. Or put down whatever you want, it's not mandatory to use a discord name ;)

If you have a discord account, we mainly communicate there: https://discord.gg/DdUt6g7gyA

PR #341, opened 2021-12-06
fix: Fix custom items without recipes

joeyparrish:
Having no requirements should imply no recipe, rather than a recipe
with no ingredients that you can make infinitely and for free.


joeyparrish:
Thanks!

sirskunkalot:
Oh, I did not answer it seems. I wrote something, though. Thank you ;)

PR #348, opened 2022-02-03
Added render cache

MSchmoecker:
A render can take 1-100ms (worst case), normally ~5-20ms if it is no complex one. This PR adds an optional `UseCache` and `TargetPlugin` option for the renderer, where sprites are written as .png to `%AppData%/LocalLow/IronGate/Valheim/Jotunn/CachedIcons`. The loading time with a cache used is 0-2ms.
The sprites are named `<prefabname>-<gameversion>.png` so they will re-render when the game version changes. When a `TargetPlugin` is set, the name is `<prefabname>-<modname>-<modversion>.png`

### DIFF
--- a/JotunnLib/Managers/RenderManager.cs
+++ b/JotunnLib/Managers/RenderManager.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using BepInEx;
 using UnityEngine;
 using Object = UnityEngine.Object;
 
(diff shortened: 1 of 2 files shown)


JotunnLib/Managers/RenderManager.cs
```diff
@@ -398,6 +461,18 @@ public class RenderRequest
...
+            ///     Optional, Used for <see cref="UseCache"/> to determine a unique name-version combination
+            /// </summary>
+            public BepInPlugin TargetPlugin { get; set; } = null;
+
+            /// <summary>
+            ///     Save the render on the disc and reuse when called again. This reduces the time to re-render drastically.
+            ///     When the game version changes, a new render will be made. When a <see cref="TargetPlugin"/> is set, the version and name
+            ///     will be used to determine if a new render should be made
+            /// </summary>
+            public bool UseCache { get; set; } = false;
```
sirskunkalot, line 474: Should be true, i think?!

PR #351, opened 2022-02-22
fix: don't error when saving unset keyboard shortcut

heinermann:
Prevents the following error message from appearing when a `KeyboardShortcut` is `KeyboardShortcut.empty` (unset) and OK is pressed:

> [Error  :   BepInEx] Failed to read keybind from settings: Requested value 'Not' was not found.


sirskunkalot:
thx, nice catch

PR #354, opened 2022-03-16
MMHOOK remove

MSchmoecker:
MMHOOK is making many problems for users, especially server owners have support questions every day as Jotunn is not running.
Very few Valheim mods use MonoMod Hooks but Harmony, so there are no major advantages. Also the stack traces are hard to read and this change will make Jotunn appear less in stack traces that it is not related to.

Harmony patches have to be static, so for Managers there will be a static subclass that is passing the patch to the Manager Instance. Not too much have to be changed this way, but some patch orders will be effected by this change and it needs to be tested with other mods when its ready

Known problems:
- ~LocalizationManager: Patches are not unsubscribed, resulting in warning `[Warning:Jotunn.Managers.LocalizationManager] Localization (com.jotunn.jotunn) already added`~
- ~InputManger Button presses sees to ignore additional keys (ctr+s already triggers with s)~


sbtoonz (approved):

sirskunkalot:
Now it's time, huh? Only problem I see is mods using MMHOOKS as well. At least PlanBuild makes heavy use of it. Not sure how many others there are as they mostly don't add HookGenPatcher as a dependency.

MSchmoecker:
Yes this is unfortunate but MMHOOK was always a dependency of Jotunn and no part of it directly. These other mods are in responsibility to mention their dependencies, even if they are transitive pulled from Jotunn

sirskunkalot:
We should also remove all the post build stuff regarding MMHOOK dlls then (JotunnBuildTask project and deployment dll)

MSchmoecker:
The JotunnBuildTask already doesn't generate MMHOOK dlls anymore and no MMHOOK assemblies are referenced inside Jotunn. With #355 there isn't a single reference to MMHOOK anymore

sbtoonz:
image
Was testing out MoarBuilds porting to new format and not sure if this is intentional but something to look at @MSchmoecker

MSchmoecker:
@sbtoonz what is the exact problem? I can see that only one category tab is active or is that from a special item from MoarBuilds? Also does this occurs in the current uploaded version of MoarBuilds?

sbtoonz:
> @sbtoonz what is the exact problem? I can see that only one category tab is active or is that from a special item from MoarBuilds? Also does this occurs in the current uploaded version of MoarBuilds?

Simply loading the mod breaks the build table for hammer 

Yields me only one category 

Going to try with a test mod and a primitive game object

MSchmoecker:
Do you have any errors? When installing MoarBuilds with this Jotunn branch it seems to work for me

sbtoonz:
> Do you have any errors? When installing MoarBuilds with this Jotunn branch it seems to work for me

It is something within  Moarbuilds
No errors , Going to look at it more this afternoon 
Using a primitive GO does not yield this outcome

sbtoonz:
> > Do you have any errors? When installing MoarBuilds with this Jotunn branch it seems to work for me
> 
> It is something within Moarbuilds No errors , Going to look at it more this afternoon Using a primitive GO does not yield this outcome

this issue was from trying to use the custom category option once that was removed from my pieces i was able to see the whole work table layout

PR #383, opened 2023-01-02
Docs: Add Character items column, fix slash in DamageModifiers list.

OrianaVenture:
- Adding a new column to the Characters list that includes all the default items and different random items given the character is a humanoid.
- Fixed the formatting for the DamageModifiers list to get rid of that extra slash in the doc.


MSchmoecker (approved):

MSchmoecker:
Thanks!

Issue #387, opened 2023-02-06
[BUG] Dropdown component not in UI layer


SpikeHimself:
Details:
Jotunn Version: 2.10.4
Jotunn Submodule(if applicable): -
Repeatability(Consistent(100%), Inconsistent(50%), Rare(1%), Unknown(==1)):  100%

**Problem Description:**
The Dropdown component of the gameobject created by GUIManager.CreateDropDown has the incorrect layer set. It is 0, where 5 (UI Layer) is expected.

**Expected Behaviour:**
    var dd = GUIManager.Instance.CreateDropDown( ... )
    dd.GetComponent\<Dropdown\>().gameObject.layer == 5

**Actual Behaviour:**
    var dd = GUIManager.Instance.CreateDropDown( ... )
    dd.GetComponent\<Dropdown\>().gameObject.layer == 0

**Link to your server's and client's LogOutput.log**
It's not visible in any logs, but a user of my mod has posted a screenshot here: https://github.com/SpikeHimself/XPortal/issues/11

SpikeHimself:
Additonal findings:

Same as above applies to the dropdown's "Template" child, i.e.

    var dd = GUIManager.Instance.CreateDropDown( ... )
    dd.transform.Find("Template").gameObject.layer == 0  // Should be 5

Changing both of these values to 5 (i.e. GUIManager.UILayer) solved the issue for my mod's user.

MSchmoecker:
Thanks for the report. It's actually interesting that it behaves like this, since the dropdown is initally created with Unity's own utility class [DefaultControls.CreateDropdown()](https://docs.unity3d.com/530/Documentation/ScriptReference/UI.DefaultControls.CreateDropdown.html) that even seems to set the layer correctly for some children.

Agreed, it makes sense that the controls are always spawned under the correct UI layer. As far as I can tell you already fixed it in your mod but feel free to open a PR if you want to provide a fix in Jotunn, too. Otherwise I will put it on my to-do list and get to it when I get to it.

PR #388, opened 2023-02-09
Fix layer for dropdown components

SpikeHimself:
Sets the correct layer for the Dropdown and ScrollRect components. Addresses #387. For more info, see https://github.com/SpikeHimself/XPortal/issues/11

### DIFF
--- a/JotunnLib/Managers/GUIManager.cs
+++ b/JotunnLib/Managers/GUIManager.cs
@@ -1362,7 +1362,10 @@ namespace Jotunn.Managers
         {
             // Move the Dropdown and ScrollRect components to the UILayer
             dropdown.gameObject.layer = UILayer;
-            dropdown.transform.Find("Template").gameObject.layer = UILayer;
+            if (dropdown.template != null)
+            {
+                dropdown.template.gameObject.layer = UILayer;
+            }
 
             // Dropdown
             if (dropdown.captionText)
--- a/JotunnLib/Managers/GUIManager.cs
+++ b/JotunnLib/Managers/GUIManager.cs
@@ -1360,6 +1360,10 @@ namespace Jotunn.Managers
         /// <param name="fontSize">Optional font size, defaults to 16</param>
         public void ApplyDropdownStyle(Dropdown dropdown, int fontSize = 16)
         {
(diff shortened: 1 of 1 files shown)


JotunnLib/Managers/GUIManager.cs
```diff
@@ -1360,6 +1360,10 @@ public void ApplyToogleStyle(Toggle toggle)
         /// <param name="fontSize">Optional font size, defaults to 16</param>
         public void ApplyDropdownStyle(Dropdown dropdown, int fontSize = 16)
         {
+            // Move the Dropdown and ScrollRect components to the UILayer
+            dropdown.gameObject.layer = UILayer;
+            dropdown.transform.Find("Template").gameObject.layer = UILayer;
```
MSchmoecker: The template child shouldn't be hardcoded since it may be named differently or be active under a second child. `dropdown.template` should work better and I think a null check for the template is also appropriate since any dropdown instance may be passed here.

MSchmoecker (changes requested):

JotunnLib/Managers/GUIManager.cs
SpikeHimself: Ah, I didn't know it had a template property. Sorry, not very familiar with Unity's UI elements. Good call on the null check, I made assumptions there.

I've pushed to the branch on my fork - should it automatically appear here? I've never done a PR before, sorry.

MSchmoecker: Yes looks good now. Yeah the changes are included here now, a PR always points to a branch :D

SpikeHimself: Splendid, sorry to be a pain! Is there anything else I need to do?

MSchmoecker: It's all good, thanks for your contribution. We all been at our first PR and this is what reviews are for :)
I'm going though a few sanity checks now, compile the changes and do a few tests to not miss something obvious and will merge it afterwards.

MSchmoecker (approved):

MSchmoecker:
Thanks for the fix

PR #391, opened 2023-03-10
Add some fonts and colours

SpikeHimself:
I wasn't sure how you'd like the font check to work going forward. Currently it checks for all existing (two) fonts: 

    if (AveriaSerifBold == null || AveriaSerif == null)
    {

It seems to me that this is enough to determine whether or not font loading has worked, and since it throws an exception I don't think it's necessary to add checks for other fonts. 

Hence I have loaded "my" fonts after the exception - at this point the first two fonts loaded fine, so we should be in the clear (right?)

        throw new Exception("Fonts not found");
    }    
    Norse = fonts.FirstOrDefault(x => x.name == "Norse");

I could move them up to before the check, so it's all in one place, but there's not really a point in loading them all and then afterwards giving up because the first two have failed. Nitpicking at this point :)

As for the colours I added, the yellow is used in various menus and headings, and the beige is found on the start screen on for example the Select World window: 

image

### DIFF
--- a/JotunnLib/Managers/GUIManager.cs
+++ b/JotunnLib/Managers/GUIManager.cs
@@ -78,9 +78,19 @@ namespace Jotunn.Managers
         public const int UILayer = 5;
 
         /// <summary>
+        ///     The default Valheim beige color.
+        /// </summary>
+        public Color ValheimBeige = new Color(0.8529f, 0.725f, 0.5331f, 1f);
+
+        /// <summary>
         ///     The default Valheim orange color.
         /// </summary>
         public Color ValheimOrange = new Color(1f, 0.631f, 0.235f, 1f);
+
+        /// <summary>
+        ///     The default Valheim yellow color.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


MSchmoecker:
Thanks for the additions.

Both colors look good, Valheim itself isn't very strict with them and uses slightly different shades all over the place. They are definitively close enough :)

The exceptions here are rather used for us to quickly see issues when Valheim updates, so the debugging time is shortened when something changes. Thus exceptions should be also added for the other fonts or at least a warning log should be printed.

SpikeHimself:
> They are definitively close enough :)

I would hope so, given that I have taken them straight out of the game itself using UnityExplorer :D

> Thus exceptions should be also added for the other fonts or at least a warning log should be printed.

Right, in that case it makes more sense to put them in a list, and iterate, or something to that extent. For now I'll just add the checks "manually".

MSchmoecker (approved):

PR #403, opened 2023-07-11
Allow mock resolution to continue when exception thrown

OrianaVenture:
Adding try-catch block to GetValue to prevent exception from failing entire mock resolution chain. Null checking already implemented in codebase for usage of this getter.


MSchmoecker:
Looks good and I haven't found a better solution for your case with TMP either. Although a try/catch should also be added to `FieldMember` so that both classes behave similarly.

MSchmoecker (approved):

PR #404, opened 2023-07-24
Fix typo in Fenring armor for Recipes

glennfu:
I was parsing this list to do a WackysDatabase export, and noticed that these two were incorrect. They're correct on the Items list, but somehow got changed in the Recipes list.


MSchmoecker:
Thanks for the contribution, however these are the correct recipe names for the current Valheim release version (0.216.9).

Maybe you are playing on the PTB and the recipe names where changed there or some other mod is changing them? I've also tried both recipes with WackysDatabase and only `Recipe_ArmorFenrirChest`/`Recipe_ArmorFenrirLegs` are accepted.

glennfu:
Just the regular latest version on Steam. I tried to export that recipe using Wackysdb and it failed. Then I noticed it didn't match the prefab name listed in Items. I tried exporting that one, and it worked.

MSchmoecker:
Please execute the commands and send your log file from BepInEx/LogOutput.log here.

PR #405, opened 2023-07-24
Add all Status Effects to data section of the documentation

OrianaVenture:
Status Effects has been left out of the documentation. Adds this section to the data section with all possible status effects from the ObjectDB.

I am unsure if I added the correct entry to the toc.yaml file to add a new tab for status effects.


MSchmoecker:
Thanks, that's much cleaner. The toc.yaml is missing an topicHref but that's an easy fix.

If you want to see the docs locally you can install docfx `dotnet tool update -g docfx` and then run `docfx JotunnLib/docfx.json --serve` in the base directory to serve the site on your computer.

PR #408, opened 2023-09-16
fix: modify GetOrAddComponent to use TryGetComponent.

redseiko:
Null-coalescing operators do not work properly on Unity objects.

See: https://stackoverflow.com/q/76294124

Also it removed the BOM for some reason.


MSchmoecker (approved):

MSchmoecker:
Thank you. Does this fix an edge case in some mods or is it a general improvement? I.e. should we push a fix for this or can it be part of the next update?

redseiko:
Just a general improvement. I'll be submitting a few more performance improvements for GUIManager as well. Can all be pushed as part of an update.

PR #415, opened 2023-11-08
OnConfigurationWindowClosed Event

searica:
I wanted to add an event that triggers when the in-game configuration manager window is closed because I realized that probablykory and I likely aren't the only ones who use the window closing as a way to trigger updating mods after configuration settings have been changed. Plus having an event for it in Jotunn would be quite convenient and reduce the amount of boilerplate code needed to set up the event in each mod I make.

Since this would be a feature that is mostly used for local management of configuration files I don't know if adding the event to the SynchronizationManager class is the most appropriate place for it or not? This was the easiest place to add the event though as the code in it already checks for the configuration manager window closing.

### DIFF
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -875,7 +875,7 @@ namespace Jotunn.Managers
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


MSchmoecker:
Thanks, looks good. Yes adding the event to the SynchronizationManager is fitting, since it's about configuration.

MSchmoecker (approved):

PR #416, opened 2023-11-17
Extensions for GameObject, Transform, and String to provide convience methods

searica:
I had originally intended to break these into a few smaller PRs but since the goal is a release this weekend I merged them into a single PR. In general each commit corresponds to a set of related extension methods for each class and explains the rationale behind adding those methods, so that it's hopefully easier to accept/reject whatever subset of extensions are judged to be appropriate for Jotunn.

### DIFF
--- a/JotunnLib/Extensions/GameObjectExtension.cs
+++ b/JotunnLib/Extensions/GameObjectExtension.cs
@@ -1,4 +1,5 @@
 using System.Reflection;
+using System;
 using TMPro;
 using UnityEngine;
 using UnityEngine.UI;
@@ -125,6 +126,42 @@ namespace Jotunn
         public static bool HasComponent(this GameObject gameObject, string componentName)
         {
             return gameObject.GetComponent(componentName) != null;
+        }
+
+        /// <summary>
+        ///     Check if GameObject has any of the specified components.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="components"></param>
+        /// <returns></returns>
+        public static bool HasAnyComponent(this GameObject gameObject, params Type[] components)
+        {
+            foreach (var compo in components)
+            {
+                if (gameObject.GetComponent(compo) != null)
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        /// <summary>
+        ///     Check if GameObject has any of the specified components.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="components"></param>
+        /// <returns></returns>
+        public static bool HasAnyComponent(this GameObject gameObject, params string[] componentNames)
+        {
+            foreach (var name in componentNames)
+            {
+                if (gameObject.GetComponent(name) != null)
+                {
+                    return true;
+                }
+            }
+            return false;
         }
     }
 
--- a/JotunnLib/Extensions/GameObjectExtension.cs
+++ b/JotunnLib/Extensions/GameObjectExtension.cs
@@ -199,6 +199,29 @@ namespace Jotunn
             }
             return true;
         }
+
+        /// <summary>
+        ///     Check if GameObject or any of it's children
+        ///     have any of the specified components.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="components"></param>
+        /// <returns></returns>
+        public static bool HasAnyComponentInChildren(
+            this GameObject gameObject,
+            bool includeInactive = false,
+            params Type[] components
+        )
+        {
+            foreach (var compo in components)
+            {
+                if (gameObject.GetComponentInChildren(compo, includeInactive) != null)
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
     }
 
     /// <summary>
--- a/JotunnLib/Extensions/GameObjectExtension.cs
+++ b/JotunnLib/Extensions/GameObjectExtension.cs
@@ -107,6 +107,7 @@ namespace Jotunn
             return target;
         }
 
+        /// <summary>
         ///     Check if GameObject has any of the specified components.
         /// </summary>
         /// <param name="gameObject"></param>
--- a/JotunnLib/Extensions/GameObjectExtension.cs
+++ b/JotunnLib/Extensions/GameObjectExtension.cs
@@ -28,7 +28,7 @@ namespace Jotunn
         /// <typeparam name="T">Any type that inherits MonoBehaviour</typeparam>
         /// <param name="this">this</param>
         /// <returns>Returns null when MonoBehaviours.op_equality returns false.</returns>
-        public static T OrNull<T>(this T @this) where T : Object
+        public static T OrNull<T>(this T @this) where T : UnityEngine.Object
         {
             return (T)(@this ? @this : null);
         }
--- a/JotunnLib/Extensions/GameObjectExtension.cs
+++ b/JotunnLib/Extensions/GameObjectExtension.cs
@@ -107,29 +107,6 @@ namespace Jotunn
             return target;
         }
 
-        /// <summary>
-        ///     Extension method to check if GameObject has a component.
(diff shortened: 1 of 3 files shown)


JotunnLib/Extensions/GameObjectExtension.cs
```diff
@@ -104,6 +106,174 @@ public static GameObject OrNull(this GameObject @this)
...
+
+        /// <summary>
+        ///     Extension method to check if GameObject has a component.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="gameObject"></param>
+        /// <returns></returns>
+        public static bool HasComponent<T>(this GameObject gameObject) where T : Component
+        {
+            return gameObject.GetComponent<T>() != null;
```
MSchmoecker: Unity actually discourages doing null checks against Unity objects (https://docs.unity3d.com/Manual/BestPracticeUnderstandingPerformanceInUnity7.html at Null comparisons against UnityEngine.Object subclasses)

Thus, the method should be simplefied to `return gameObject.GetComponent<T>()`, at which point the whole method is rather redundent and doesn't have a purpose anymore. It should be removed in my opinion, same for the string overload.

JotunnLib/Extensions/GameObjectExtension.cs
```diff
@@ -104,6 +106,174 @@ public static GameObject OrNull(this GameObject @this)
...
+        ///     Check if GameObject has any of the specified components.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="components"></param>
+        /// <returns></returns>
+        public static bool HasAnyComponent(this GameObject gameObject, params Type[] components)
+        {
+            foreach (var compo in components)
+            {
+                if (gameObject.GetComponent(compo) != null)
```
MSchmoecker:
```suggestion
                if (gameObject.GetComponent(compo))
```

because of the mentioned null checks. This also applies to the follwing methods.

JotunnLib/Extensions/GameObjectExtension.cs
```diff
@@ -104,6 +106,174 @@ public static GameObject OrNull(this GameObject @this)
...
+            return true;
+        }
+
+        /// <summary>
+        ///     Check if GameObject or any of it's children
+        ///     have any of the specified components.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="components"></param>
+        /// <returns></returns>
```
MSchmoecker: The includeInactive paramter is missing in the docs

JotunnLib/Extensions/GameObjectExtension.cs
```diff
@@ -104,6 +106,174 @@ public static GameObject OrNull(this GameObject @this)
...
+        /// <summary>
+        ///     Extension method to get the first component in the GameObject
+        ///     or it's children that has the specified name.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="gameObject"></param>
+        /// <param name="name"></param>
+        /// <param name="includeInactive">Whether to include inactive child objects in the search or not.</param>
+        /// <returns></returns>
+        public static T GetComponentInChildrenByName<T>(
```
MSchmoecker: I'm not sure what for the use case for this method is, why is the component serched by T and checked with a name?

JotunnLib/Extensions/GameObjectExtension.cs
```diff
@@ -104,6 +106,174 @@ public static GameObject OrNull(this GameObject @this)
...
+        ///     a breadth-first or depth-first search. Default is breadth-first.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="childName">Name of the child object to search for.</param>
+        /// <param name="breadthFirst"> Whether to preform a breadth first or depth first search. Default is breadth first.</param>
+        /// <returns></returns>
+        public static Transform FindDeepChild(this GameObject gameObject, string childName, bool breadthFirst = true)
+        {
+            return gameObject?.transform.FindDeepChild(childName, breadthFirst);
+        }
```
MSchmoecker: Valheim already has the method `Utils.FindChild`, which does the same. Although Wrapping this into an extension method could still be done, it should call the vanilla method.

Also, the null conditional (.?) on the gameobect doesn't work as expedted because of the Unity lifecyle [(it can return true despite the object being destroyed)](https://stackoverflow.com/questions/62678228/why-does-c-sharp-null-conditional-operator-not-work-with-unity-serializable-vari) and should be removed anyway. It's not the purose if this method to test if the gameObject is null.

JotunnLib/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static string EmptyIfNull(this object value)
+        {
+            if (value == null)
+            {
+                return string.Empty;
+            }
+            return value.ToString();
+        }
```
MSchmoecker: This method is redundent and can be written as `string ?? string.Empty` or `string ?? ""`

JotunnLib/Extensions/TransformExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        /// <summary>
+        ///     Extension method to find nested children by name using either
+        ///     a breadth-first or depth-first search. Default is breadth-first.
+        /// </summary>
+        /// <param name="transform"></param>
+        /// <param name="childName">Name of the child object to search for.</param>
+        /// <param name="breadthFirst"> Whether to preform a breadth first or depth first search. Default is breadth first.</param>
+        /// <returns></returns>
+        public static Transform FindDeepChild(
```
MSchmoecker, line 18: As mentioned above, this method is already implemented in `Utils.FindChild` and should call the vanilla method

MSchmoecker (changes requested):

JotunnLib/Extensions/GameObjectExtension.cs
searica: Oh, that's very good to know. Thank you! Will remove that then.

searica: And also update some of my own code in other projects now that I know that

JotunnLib/Extensions/GameObjectExtension.cs
searica: I made it because there are a few cases in MoreVanillaBuildPrefabs where I need to search for a Component in a specific child object but I may not know what the transform tree looks like.

 I think I wrote it before I figured out how to search for a specific nested child or that Utils.FindChild existed. I'll update the PR to remove this one since it's better accomplished by using Utils.FindChild(name).GetComponent<T>() instead.

JotunnLib/Extensions/GameObjectExtension.cs
searica: Thanks for letting me know, will update accordingly and also very much appreciate the link to help learn more about Unity (will make sure to account for that in my own code now).

JotunnLib/Extensions/StringExtensions.cs
searica: Yep, I wrote that before I learned about the ?? operator in C#. I'll remove this one as well.

JotunnLib/Extensions/GameObjectExtension.cs
searica: Thanks for catching that, will fix that now!

searica:
I've pushed a new set of changes that hopefully incorporate the feedback you gave. I'm not sure that the way I rewrote HasAllComponents to avoid null checks is the best way though. 

Edit: I did some more reading on MonoBehavior and how Unity handles implicitly casting them to a bool, so I removed the counter in HasAllComponents to optimize it a bit more.

MSchmoecker (approved):
Looks good, thanks for the extensions

PR #417, opened 2023-12-19
Added feature to help resolve mocks based off same type search.

OrianaVenture:
* Accepts new mock format JVLmock_PrefabName__ChildName.
* Finds an object that matches the type of the mock from the specified child of the prefab in the new format.
* Cleans name of any object Mesh with " Instance" on the end of the name for finding mock replacement.

Since there are many unnamed Mesh objects used the current mocking system can not replace them. This is the required changes to get a functioning mesh mocking system, but should work for other types too.

Examples of usage:
1. Create a mocked mesh called JVLmock_Pickable_Thistle__default, this will find the first Mesh on the Pickable_Thistle prefab under the child "default" for a replacement.
2. Create a mocked mesh named JVLmock_TrollCave02__dirtfloor (11), this will find the first Mesh on the TrollCave02 prefab under the child "dirtfloor (11)" for a replacement.

### DIFF
--- a/JotunnLib/Managers/MockSystem/MockManager.cs
+++ b/JotunnLib/Managers/MockSystem/MockManager.cs
@@ -142,76 +142,56 @@ namespace Jotunn.Managers
         /// <returns>the real prefab</returns>
         public static Object GetRealPrefabFromMock(Object unityObject, Type mockObjectType)
         {
-            if (!unityObject)
+            if (!unityObject || !IsMockName(unityObject.name, out string assetName, out List<string> childNames))
             {
                 return null;
             }
 
-            var unityObjectName = unityObject.name;
-            if (IsMockName(unityObjectName, out unityObjectName, out List<string> childNames))
+            if (childNames.Count == 0 && TryGetAsset(mockObjectType, assetName, out Object asset))
             {
-                if (childNames != null && childNames.Count > 0)
-                {
-                    // Handle mocks that require path of existing prefab to find/replace
-                    // These are represented by JVLMock_PrefabName__ChildName__ChildName2 etc
-
-                    childNames[childNames.Count - 1] = GetCleanedName(mockObjectType, childNames[childNames.Count - 1]);
-
-                    var parent = PrefabManager.Cache.GetPrefab<GameObject>(unityObjectName);
-
-                    if (!parent)
-                    {
-                        var path = string.Join<string>("/", childNames);
-                        throw new MockResolveException(
-                            $"Mock {mockObjectType.Name} at '{unityObjectName}' with child path '{path}' could not be resolved. Prefab '{unityObjectName}' not found.",
-                            unityObjectName, path, mockObjectType);
-                    }
-
-                    var child = parent.FindDeepChild(childNames);
-
-                    if (!child || child.name != childNames.Last())
-                    {
-                        var path = string.Join<string>("/", childNames);
-                        throw new MockResolveException(
-                            $"Mock {mockObjectType.Name} at '{unityObjectName}' with child path '{path}' could not be resolved. Child '{childNames.Last()}' not found with the specified path.",
-                            unityObjectName, path, mockObjectType);
-                    }
-
-                    var obj = FindObjectInChildren(child.gameObject, mockObjectType);
-
-                    if (obj == null)
-                    {
-                        var path = string.Join<string>("/", childNames);
-                        var usedPath = child.GetPath().TrimStart('/');
-                        throw new MockResolveException(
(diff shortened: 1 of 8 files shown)


MSchmoecker:
Thanks, it looks really good. Especially using the `ClassMember` to find fields on the child, which makes it much more generic.
Using `__` as a separator instead of a single underscore is also a good idea to keep the names intact and feels natural to use.

While testing with the `VikingShip`, a child name can be ambiguous, e.g. `mast` is used for both the mesh and the interactive child. The interactive child is nested shallower and therefore found first. Intermediate children could be a solution, so maybe `JVLmock_VikingShip__ship__mast` to match the `ship` child first and then the `mast` child.

With more complex matching, the warnings should also be more specific. For example, if a child wasn't found should print a different warning than if the type on the child wasn't found. If you don't want to worry about this, I can also do this later.

There are a few minor comments I will make on the code but this won't be a blocker.

JotunnLib/Managers/MockSystem/MockManager.cs
```diff
@@ -142,22 +147,40 @@ public static Object GetRealPrefabFromMock(Object unityObject, Type mockObjectTy
...
+                if (pathName != null)
                 {
-                    unityObjectName = unityObjectName.RemoveSuffix(" (Instance)");
-                }
+                    // Handle mocks that require path of existing prefab to find/replace
+                    // These are represented by JVLMock_PrefabName__ChildName
+                    pathName = GetCleanedName(mockObjectType, pathName);
 
-                Object ret = PrefabManager.Cache.GetPrefab(mockObjectType, unityObjectName);
+                    GameObject parent = PrefabManager.Cache.GetPrefab(typeof(GameObject), unityObjectName) as GameObject;
```
MSchmoecker:
```suggestion
                    GameObject parent = PrefabManager.Cache.GetPrefab<GameObject>(unityObjectName);
```

JotunnLib/Managers/MockSystem/MockManager.cs
```diff
@@ -142,22 +147,40 @@ public static Object GetRealPrefabFromMock(Object unityObject, Type mockObjectTy
...
+                    var childTransform = parent?.FindDeepChild(pathName);
 
-                if (!ret)
-                {
-                    throw new MockResolveException($"Mock {mockObjectType.Name} {unityObjectName} could not be resolved", unityObjectName, mockObjectType);
+                    var obj = FindObjectInChildren(childTransform?.gameObject, mockObjectType);
```
MSchmoecker: `?.` should not be used on Unity types because it bypasses the Unity lifecycle. Although with more specifc error messages, null cases will likely be catched anyway. So `?.` won't be necessary anymore

JotunnLib/Managers/MockSystem/MockManager.cs
```diff
@@ -142,22 +147,40 @@ public static Object GetRealPrefabFromMock(Object unityObject, Type mockObjectTy
...
 
             var unityObjectName = unityObject.name;
-            if (IsMockName(unityObjectName, out unityObjectName))
+            if (IsMockName(unityObjectName, out unityObjectName, out string pathName))
             {
-                // Cut off the suffix in the name to correctly query the original material
-                if (unityObject is Material)
+                unityObjectName = GetCleanedName(unityObject.GetType(), unityObjectName);
+
+                if (pathName != null)
```
MSchmoecker:
```suggestion
                if (!string.IsNullOrEmpty(pathName))
```
This is a bit cleaner, as an empty string cannot be matched as well

JotunnLib/Managers/MockSystem/MockManager.cs
```diff
@@ -186,11 +209,24 @@ internal static void FixReferences(object objectToFix, int depth)
...
 
-        private static bool IsMockName(string name, out string cleanedName)
+        private static bool IsMockName(string name, out string cleanedName, out string pathName)
         {
+            pathName = null;
+
             if (name.StartsWith(JVLMockPrefix, StringComparison.Ordinal))
             {
-                cleanedName = name.Substring(JVLMockPrefix.Length);
+                int separator = name.IndexOf(JVLMockSeparator);
```
MSchmoecker:
```suggestion
                int separator = name.IndexOf(JVLMockSeparator, StringComparison.Ordinal);
```
using an Ordinal comparision is way faster and culture specifc strings are not relevant here

OrianaVenture:
Current minor suggestions have been implemented. I will look into making the pathing more dynamic to accept more nesting for cases like: JVLmock_VikingShip__ship__mast and work on better error logging for that case.

OrianaVenture:
Testing this on my current mocked locations the startup time is significantly slower than my stand alone solution which had caching. Next improvement will be to add in the ability to cache all instances with the longer path finding names after finding them once.

MSchmoecker (approved):

PR #418, opened 2023-12-23
Added  pre-defined BeInExPath to be considered in JotunnLibRefsCorlib…

Defee:
Hi the reason for this is to improve the flexibility when building with the Environment.props or via setting up different place where BepInEx resides. It gives a bit more flexibility to the developer and allows to use mode managers like Vortex, Thunderstore as bepinex base when building.


MSchmoecker:
Thanks, for consistency the env variable should be named in caps and the PropertyGroup could be moved to Paths.props, so it's possible to reuse it when needed.

MSchmoecker (approved):

PR #419, opened 2023-12-28
Update localization.md

jneb802:
Clarified syntax required for loading text assets from asset bundles for localization languages.


MSchmoecker:
The image should not be hosted on an external provider like imgur but instead placed as a file in the repo and referenced via a relative path. Otherwise looks good.

PR #425, opened 2024-03-24
RenderManager: parse dependencies of subtype

bid-soft:
As discussed on discord. Parsing subtype dependencies for RenderManager.
I have not been able to test it, as I was not able to reproduce the original error with ?jewelcrafting?.


MSchmoecker (approved):
Looks good and can confirm the original issue is fixed, too. The setup is a bit complicated, you'd need VNEI, Jewelcrafting and clear the icon cache so a re render is triggered.

PR #427, opened 2024-04-27
Fix for 2 NREs related to assets

probablykory:
Fix for 2 NREs that seem to occur while running Jotunn mods alongside Krumpac's Reforged modpack.

### DIFF
--- a/JotunnLib/Managers/PrefabManager.cs
+++ b/JotunnLib/Managers/PrefabManager.cs
@@ -441,7 +441,7 @@ namespace Jotunn.Managers
                         // load an never release reference to keep it loaded
                         asset.Load();
 
-                        if (asset.Asset.GetType() == type)
+                        if (asset.Asset != null && asset.Asset.GetType() == type)
                         {
                             return asset.Asset;
                         }
@@ -453,7 +453,8 @@ namespace Jotunn.Managers
                     }
                 }
 
-                if (GetCachedMap(type).TryGetValue(name, out var unityObject))
+                var cacheMap = GetCachedMap(type);
+                if (cacheMap != null && cacheMap.TryGetValue(name, out var unityObject))
                 {
                     return unityObject;
                 }
--- a/JotunnLib/Managers/PrefabManager.cs
+++ b/JotunnLib/Managers/PrefabManager.cs
@@ -441,20 +441,22 @@ namespace Jotunn.Managers
                         // load an never release reference to keep it loaded
                         asset.Load();
 
-                        if (asset.Asset != null && asset.Asset.GetType() == type)
+                        if (asset.Asset)
                         {
-                            return asset.Asset;
-                        }
+                            if (asset.Asset.GetType() == type)
+                            {
+                                return asset.Asset;
+                            }
 
-                        if (asset.Asset is GameObject gameObject && TryFindAssetInSelfOrChildComponents(gameObject, type, out Object childAsset))
-                        {
-                            return childAsset;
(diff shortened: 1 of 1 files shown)


JotunnLib/Managers/PrefabManager.cs
```diff
@@ -441,7 +441,7 @@ public static Object GetPrefab(Type type, string name)
                         // load an never release reference to keep it loaded
                         asset.Load();
 
-                        if (asset.Asset.GetType() == type)
+                        if (asset.Asset != null && asset.Asset.GetType() == type)
```
MSchmoecker: If a null check is added here, the statement below has to be covered too. This rather points to Reforged adding an asset that doesn't live anymore but a check definitely won't hurt. Although it's unexpected since asset.IsValid returns true, best to ckeck how Reforged is adding assets into the vanilla sytem.

`asset.Asset &&` should rather be used than `asset.Asset != null &&`, because the first one is checking for Unity liveness and is more reliable and much cheaper.

JotunnLib/Managers/PrefabManager.cs
```diff
@@ -453,7 +453,8 @@ public static Object GetPrefab(Type type, string name)
                     }
                 }
 
-                if (GetCachedMap(type).TryGetValue(name, out var unityObject))
+                var cacheMap = GetCachedMap(type);
+                if (cacheMap != null && cacheMap.TryGetValue(name, out var unityObject))
```
MSchmoecker: GetCachedMap cannot and should not return null in any case, can you explain what's the reasoning here?

JotunnLib/Managers/PrefabManager.cs
probablykory: Makes sense.  I'll make some adjustments.

JotunnLib/Managers/PrefabManager.cs
probablykory: I obviously hadn't looked too far into GetCachedMap.  Part of my problem resolving my issue is that Reforged is huge so takes forever to test after making changes.  Additionally it is protected by keymanager, thus throws hard when a debugger is attached.

This PR was simply my attempt to prevent any possible NRE from happening within GetPrefab.

I'll take this change out when updating the PR.

probablykory:
Tested normal operation (0.217.46 with MoreCrossbows and Jotunn), and this change seems to be benign.  Tested with Reforged, and with the change it works fine.  Without the change (eg tested with Jotunn 2.19.3 to verify) it does indeed throw an NRE.

```
[Error  :Jotunn.Managers.GUIManager] System.NullReferenceException: Object reference not set to an instance of an object
  at Jotunn.Managers.PrefabManager+Cache.GetPrefab (System.Type type, System.String name) [0x0002a] in D:\Projects\Jotunn\JotunnLib\Managers\PrefabManager.cs:444
  at Jotunn.Managers.PrefabManager+Cache.GetPrefab[T] (System.String name) [0x00000] in D:\Projects\Jotunn\JotunnLib\Managers\PrefabManager.cs:472
  at Jotunn.Managers.GUIManager.InitializeAssets () [0x002ec] in D:\Projects\Jotunn\JotunnLib\Managers\GUIManager.cs:423
```
Full log [attached](https://github.com/Valheim-Modding/Jotunn/files/15155930/LogOutput.log).

MSchmoecker:
thanks, looks good now

MSchmoecker (approved):

PR #429, opened 2024-05-04
Remove JotunnBuildTask, Add BepInEx.AssemblyPublicizer, Switch projects to modern SDK style

MSchmoecker:


MSchmoecker:
Not merging because BepInEx.AssemblyPublicizer only accepts modern SDK style projects. This would cause compile errors for mods that used the template/.NetFramework and is not worth the trouble.

PR #430, opened 2024-05-10
DungeonManager

probablykory:
Adding a new DungeonManager to Jotunn, allowing devs to add custom dungeon rooms to Valheim.


JotunnLib/Managers/DungeonManager.cs
```diff
@@ -0,0 +1,387 @@
...
+        {
+            if (assetBundle == null)
+            {
+                throw new ArgumentException("Cannot be null", nameof(assetBundle));
+            }
+            if (string.IsNullOrEmpty(prefabName))
+            {
+                throw new ArgumentException("Cannot be null or empty", nameof(prefabName));
+            }
+            envNames.Add(prefabName, assetBundle);
```
MSchmoecker: prefabs should be loaded early when possible, mostly to give direct feedback if a prefab is missing or an AssetBundle is invalid.

JotunnLib/Managers/DungeonManager.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        /// <summary>
+        ///     Adds a new prefab to be registered when <see cref="ZoneSystem.SetupLocations"/> runs.<br />
+        ///     If you intend to use a custom door, this method enables you provide that prefab here.
+        /// </summary>
+        /// <param name="assetBundle">The <see cref="AssetBundle"/> containing the prefab.</param>
+        /// <param name="prefabName">The name of the prefab to register.</param>
+        public void RegisterPrefab(AssetBundle assetBundle, string prefabName)
```
MSchmoecker: What's the reason to have a dedicated method here instead of using `PrefabManager.Instance.AddPrefab` directly? Especially if the prefab has a ZNetView, `ZoneSystem.SetupLocations` might be too late

JotunnLib/Configs/RoomConfig.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        /// <summary>
+        ///     Theme name of this room.
+        /// </summary>
+        public string ThemeName { get; }
+
+        /// <summary>
+        ///     <see cref="Room.Theme"/> of this room.
+        /// </summary>
+        public Room.Theme Theme { get; }
```
MSchmoecker: Instead of having both ThemeName and Theme, it would be more suitable to only use the string externally and convert internally to `Room.Theme` when needed

JotunnLib/Entities/CustomRoom.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        ///     Custom room from a prefab loaded from an <see cref="AssetBundle"/> with a <see cref="global::Room"/> made from a <see cref="RoomConfig"/>.<br />
+        ///     Can fix references for <see cref="Entities.Mock{T}"/>s.
+        /// </summary>
+        /// <param name="assetBundle">A preloaded <see cref="AssetBundle"/></param>
+        /// <param name="assetName">Name of the prefab in the bundle.</param>
+        /// <param name="roomConfig">The config for this custom room.</param>
+        public CustomRoom(AssetBundle assetBundle, string assetName, RoomConfig roomConfig)
+            : this(assetBundle, assetName, false, roomConfig) { }
```
MSchmoecker: Constructors without a `fixReference` are considered obsolete and should not be provided in new interfaces for consistency

JotunnLib/Configs/RoomConfig.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        ///     Flag to ensure this room is only placed around the perimeter of a camp.  Used only for generating camps (draugr/fuling villages).
+        /// </summary>
+        public bool? Perimeter { get; set; }
+
+        /// <summary>
+        ///     Background music to play when a player gets in range of this room.
+        /// </summary>
+        public MusicVolume MusicPrefab { get; set; }
```
MSchmoecker: Config classes must be (JSON) serializable, thus a direct Unity Object reference cannot be used

JotunnLib/Entities/CustomRoom.cs
```diff
@@ -0,0 +1,172 @@
...
+            {
+                m_prefab = new SoftReference<GameObject>(AssetManager.Instance.AddAsset(Prefab)),
+                m_loadedRoom = Room,
+                m_prefabData = new RoomPrefabData()
+                {
+                    m_enabled = Room.m_enabled,
+                    m_theme = Theme
+                }
+            };
+        }
```
MSchmoecker: Apply methods are usually located in the config classes, not the manager directly. For consistency this should be moved.

JotunnLib/Managers/DungeonManager.cs
```diff
@@ -0,0 +1,387 @@
...
+        ///     The singleton instance of this manager.
+        /// </summary>
+        public static DungeonManager Instance => _instance ??= new DungeonManager();
+
+        /// <summary>
+        ///     Event that gets fired after the vanilla <see cref="DungeonDB"/> has loaded rooms.  Your code will execute
+        ///     every time a main scene is started (on joining a game). <br />
+        ///     If you want to execute just once you will need to unregister from the event after execution.
+        /// </summary>
+        public static event Action OnVanillaRoomsAvailable;
```
MSchmoecker, line 29: Best to provide a OnRoomsRegistered event, too

JotunnLib/Managers/DungeonManager.cs
```diff
@@ -0,0 +1,387 @@
...
+            {
+                DungeonDB.RoomData result = Instance.OnDungeonDBGetRoom(hash);
+
+                if (result != null)
+                {
+                    __result = result;
+                    return false;
+                }
+                return true;
+            }
```
MSchmoecker: The room is already `DungeonDB.instance.m_rooms` and the `DungeonDB.instance.m_roomByHash` is regenerated, so all this extra handling doesn't look necessary

JotunnLib/Configs/RoomConfig.cs
```diff
@@ -0,0 +1,92 @@
+using Jotunn.Entities;
+
+namespace Jotunn.Configs
+{
+    /// <summary>
+    ///     Configuration class for adding custom clutter.<br />
```
MSchmoecker: you missed to replace clutter :P

MSchmoecker (commented):
This are my first comments reading the code, overall pretty good.

The DungeonGeneratorTheme helper looks fine too me.

JotunnLib/Managers/DungeonManager.cs
probablykory: This was originally written before the new asset system, and if dungeon prefabs used here used mocks of materials/sounds/etc from vanilla dungeon parts, they would fail to resolve when using PrefabManager.Instance.AddPrefab.

In general, it's meant to register prefabs that are specifically used by or referenced from the dungeon generator.

The point about znetview is interesting.

I'll dig into this at some point soon - if mocks referencing dungeon resources now work due to the new asset system, this may not be needed at all anymore.

JotunnLib/Configs/RoomConfig.cs
probablykory: I kept both in order to allow devs to add additional rooms to existing dungeons.  Since Room.Theme is always needed (either with one of the vanilla values, or 0 for None), it made sense to me to make themeName optional instead.

I could also do this and hide the Theme behind a string parse, and then when applying the themeName, attempt to parse the string as the enum, but at the time it seemed more complicated and unnecessary.

JotunnLib/Configs/RoomConfig.cs
probablykory: That makes sense.  I'll take this out for now, and work out how to provide that option again down the line.  I wanted parity with features available in the editor.

JotunnLib/Managers/DungeonManager.cs
```diff
@@ -0,0 +1,387 @@
...
+                }
+                else
+                {
+                    Logger.LogDebug($"Adding additional rooms of type {self.m_themes} to available rooms");
+
+                    DungeonGenerator.m_availableRooms.AddRange(Rooms.Values
+                        .Where(r => r.Room.m_enabled)
+                        .Where(r => self.m_themes.HasFlag(r.Theme))
+                        .Select(r => r.RoomData));
+                }
```
MSchmoecker: Doesn't this branch potentially adds rooms twice to the `m_availableRooms` list, because it's already populated by the vanilla call?

JotunnLib/Configs/RoomConfig.cs
MSchmoecker: A prefab name would be fine at this point, so it can be resolved with the mock system

JotunnLib/Managers/DungeonManager.cs
probablykory: The extra handling is necessary because `DungeonDB.instance.m_rooms` and `DungeonDB.instance.m_roomByHash` both do not contain any rooms with a Theme value of 0.  By default DungeonGenerator omits any rooms marked with either Theme=0 or Enabled=False.  Rather than patching the Generator to change the necessary logic (potential future maintenance headaches), it keeps it's own list of custom rooms and patches both `DungeonGenerator.SetupAvailableRooms` and `DungeonDB.GetRoom` to ensure the custom rooms are available for fully custom dungeons or for custom rooms for vanilla dungeons.

JotunnLib/Managers/DungeonManager.cs
probablykory: See explanation in this [comment](https://github.com/Valheim-Modding/Jotunn/pull/430#discussion_r1597048794) about how this works.

JotunnLib/Managers/DungeonManager.cs
MSchmoecker: I see, missed the theme handling.
However, it seems undesirable to have `DungeonDB.GetRooms()` and `DungeonDB.GetRoom(hash)` behave differently. In my opinion, it would be cleaner to either adding the rooms to the vanilla system and patching `SetupAvailableRooms` or not adding them to the vanilla system at all and only patching all necessary methods.

probablykory: Yeah that argument makes sense.  I think there was more of a necessity for the blended approach back when DungeonDB did more setup, before they added the soft reference system.

I'll pull out the logic related to and using `RegisterRoomInDungeonDB`, and see if there are any functional gaps that need to be patched.

JotunnLib/Managers/DungeonManager.cs
probablykory: I tested out using regular ol `PrefabManager.Instance.AddPrefab` and it seems to work for my purposes now.  Iirc, the issue I had prior to 0.217.46 was that mocks for dungeon resources would not resolve.  I'll be removing the RegisterPrefab method + logic in my next PR update.

JotunnLib/Managers/DungeonManager.cs
jneb802: @MSchmoecker Do I need to do anything other than just adding the below line?

`public static event Action OnRoomsRegistered;`

MSchmoecker: the event has to be invoked like `OnVanillaRoomsAvailable` with `InvokeOnVanillaRoomsAvailable` after the rooms have been registered

JotunnLib/Managers/DungeonManager.cs
jneb802: @MSchmoecker Could you further clarify what needs to change here? I'm not sure what I should change in this code from your comment.

MSchmoecker: The AssetBundle is loaded at later time than the `RegisterEnvironment` call: https://github.com/Valheim-Modding/Jotunn/blob/24ce92e2ccb448652ea7e449cded4fd1ccc923f8/JotunnLib/Managers/DungeonManager.cs#L257-L267

The loading part should be moved to RegisterEnvironment, while the `env.FixReferences(true)` should stay at OnZoneSystemSetupLocations to make sure vanilla assets are loaded.

Looking at this code part, `UnityEngine.Object.Instantiate<GameObject>(env)` should also changed to spawn under a `DontDestroyOnLoad` object to not be destroyed on scene change. See https://github.com/Valheim-Modding/Jotunn/blob/294c8304f3d6ff1270420eb64a395be98aa8b16c/JotunnLib/Managers/PrefabManager.cs#L69-L71 and related code

JotunnLib/Managers/DungeonManager.cs
jneb802: @MSchmoecker I'm also struggling with this comment. What am I supposed to change here?

MSchmoecker: There's no clear suggestion to make here. It should work and behave consistently if possible, ie. the current implementation works but isn't consistent. If you don't have a better way, you can leave it as it is

MSchmoecker: maybe the room should just be added to `DungeonDB.m_rooms` and `DungeonDB.m_roomByHash`. Then there isn't a need for the patch and both methods behave consistently

JotunnLib/Managers/DungeonManager.cs
jneb802: Addressed in: https://github.com/Valheim-Modding/Jotunn/pull/434/commits/9338dc963c4127a7e81e3f7f8bdfc267c6ffe323

JotunnLib/Managers/DungeonManager.cs
jneb802: Addressed in https://github.com/Valheim-Modding/Jotunn/pull/434/commits/f515abe679c8ca82112bb2ff4c435246aca2ed8d

JotunnLib/Entities/CustomRoom.cs
jneb802: Addressed in: https://github.com/Valheim-Modding/Jotunn/pull/434/commits/5fb11c61edd45d61c090db6a73e9967e53e0c3f0

JotunnLib/Configs/RoomConfig.cs
jneb802: Addressed in: https://github.com/Valheim-Modding/Jotunn/pull/434/commits/366b7ab3e91eaced3022366f83654e742578c06b

JotunnLib/Entities/CustomRoom.cs
jneb802: Addressed in: https://github.com/Valheim-Modding/Jotunn/pull/434/commits/d218a4db11c0522c22ad899e2ead3ec4e5c855f4

JotunnLib/Managers/DungeonManager.cs
jneb802: Addressed in: https://github.com/Valheim-Modding/Jotunn/pull/434/commits/611bdfd2bb0fe75b69950e0b8df0c224834ba0f3

JotunnLib/Configs/RoomConfig.cs
jneb802: Addressed in: https://github.com/Valheim-Modding/Jotunn/pull/434/commits/5991d7793bc68f46effaa47970290fd1f5e6f39f

probablykory:
Rebased my branch to latest dev & Cherrypicked warp's PR changes.  I've got more changes pending, but wanted to get these up.

JotunnLib/Managers/DungeonManager.cs
```diff
@@ -251,7 +256,7 @@ private void OnZoneSystemSetupLocations()
 
                     var env = kvp.Value;
                     env.FixReferences(true);
-                    UnityEngine.Object.Instantiate<GameObject>(env, DungeonRoomContainer.transform);
+                    UnityEngine.Object.Instantiate<GameObject>(env); // Intentionally omitting DungeonRoomContainer.transform
```
probablykory, line 256: The env gameobject needs to be enabled in order for the component awakes to run.

JotunnLib/Configs/RoomConfig.cs
probablykory: Fixed in 788010f7c6958a35aca8ec43c8579491b8a63ab4

JotunnLib/Entities/CustomRoom.cs
probablykory: Fixed for real in 47472f1df77d941d7f7afbb86f56e21ad6cfc35e

MSchmoecker (approved):

PR #441, opened 2024-11-03
Add SynchronizationModeAttribute to allow mod authors to control how AdminOnly configs are handled.

searica:
This adds a new attribute that can be applied to the base plugin which is checked by `SynchronizationManager` during `SynchronizationManager.Instance.GetConfigFiles()` to determine if each config file should be handled by `SynchronizationManager` or not based on the SynchronizationModeAttribute and if Jotunn is installed on the server. 

If mod authors do not set the SynchronizationModeAttribute then it defaults to the current behaviour (which is the same as explicitly setting `EnforceAdminOnly = AdminOnlyStrictness.Always` ). 

If mod authors set or set `EnforceAdminOnly = AdminOnlyStrictness.IfJotunnOnServer` then when Jotunn is not present on the server SynchronizationManager will not manage the config file at all and the mod will behave as if there is no synchronization and it is running as a purely client side mod. If Jotunn is present though it will enforce the AdminOnly restriction and sync all configuration changes.

I've tested it with existing mods and those tests suggest these features are backwards compatible and preserve the current configuration behaviour when used with mods built against the currently live version of Jotunn. I've also written the code to handle mapping custom config files to their related plugins so that the SynchronizationModeAttribute can still be checked.

Please let me know if you'd like any additional testing to be done and I'm happy to hear any other feedback or suggestions.

### DIFF
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -498,10 +498,6 @@ namespace Jotunn.Managers
         /// <returns></returns>
         private bool ShouldManageConfig(ConfigFile config)
         {
-            if (ModCompatibility.IsJotunnOnServer())
-            {
-                return true;
-            }
             if (!GetPluginGUID(config, out var pluginGUID))
             {
                 return false;
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -874,6 +874,11 @@ namespace Jotunn.Managers
         {
             foreach (var config in GetConfigFiles())
             {
+                if (!ShouldManageConfig(config))
+                {
+                    continue;
+                }
+
                 foreach (var configDefinition in config.Keys)
                 {
                     var configEntry = config[configDefinition.Section, configDefinition.Key];
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -473,11 +473,12 @@ namespace Jotunn.Managers
         {
             var loadedPlugins = BepInExUtils.GetDependentPlugins(true);
 
-            foreach (BaseUnityPlugin plugin in loadedPlugins.Values)
+            foreach (var plugin in loadedPlugins.Values)
             {
                 yield return plugin.Config;
             }
-            foreach (ConfigFile customConfigFile in CustomConfigs.Values)
+
+            foreach (var customConfigFile in CustomConfigs.Values)
             {
                 yield return customConfigFile;
             }
@@ -495,10 +496,12 @@ namespace Jotunn.Managers
             {
                 return false;
             }
+
             if (!BepInExUtils.GetDependentPlugins().TryGetValue(pluginGUID, out var plugin))
             {
                 return false;
             }
+
             return ShouldManageConfig(plugin);
         }
 
@@ -520,11 +523,7 @@ namespace Jotunn.Managers
             // not been set for the mod then return true to mimic current behaviour and 
             // maintain backwards compatibility.
             SynchronizationModeAttribute syncMode = plugin.GetSynchronizationModeAttribute();
-            if (syncMode == null || syncMode.ShouldAlwaysEnforceAdminOnly())
-            {
-                return true;
-            }
-            return false;
+            return syncMode == null || syncMode.ShouldAlwaysEnforceAdminOnly();
         }
 
         private static string GetFileIdentifier(ConfigFile config)
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -466,8 +466,7 @@ namespace Jotunn.Managers
         }
 
         /// <summary>
-        ///     Gets an IEnumerable of all default and custom config files that should be managed by SynchronizationManager. 
-        ///     Ignores config files for plugins with AdminOnlyStrictness == Low if Jotunn is not installed on the server.
+        ///     Gets an IEnumerable of all default and custom config files that associated with plugins that have Jotunn as a dependency.
         /// </summary>
         /// <returns></returns>
         private IEnumerable<ConfigFile> GetConfigFiles()
@@ -476,17 +475,11 @@ namespace Jotunn.Managers
 
             foreach (BaseUnityPlugin plugin in loadedPlugins.Values)
             {
-                if (ShouldManageConfig(plugin))
-                {
-                    yield return plugin.Config;
-                }
+                yield return plugin.Config;
             }
             foreach (ConfigFile customConfigFile in CustomConfigs.Values)
             {
-                if (ShouldManageConfig(customConfigFile))
-                {
-                    yield return customConfigFile;
-                }
+                yield return customConfigFile;
             }
         }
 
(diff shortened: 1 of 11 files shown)


JotunnLib/Utils/ModCompatibility/ModCompatibility.cs
```diff
@@ -30,6 +31,15 @@ internal static void Init()
             Main.Harmony.PatchAll(typeof(ModCompatibility));
         }
 
+        public static bool IsJotunnOnServer()
+        {
+            if (ZNet.instance && ZNet.instance.IsClientInstance())
+            {
+                return LastServerVersion != null;
```
MSchmoecker: The `LastServerVersion` can be set, even if no mods are installed on the server. This check also has to lookup if Jotunn is installed.
Probably easiest if `LastServerVersion` is changed to be a `ModuleVersionData`

JotunnLib/Managers/SynchronizationManager.cs
```diff
@@ -464,26 +465,102 @@ private void InvokeOnAdminStatusChanged()
...
+                    yield return plugin.Config;
+                }
+            }
+            foreach (ConfigFile customConfigFile in CustomConfigs.Values)
+            {
+                if (ShouldManageConfig(customConfigFile))
+                {
+                    yield return customConfigFile;
+                }
+            }
```
MSchmoecker: Have you checked that the timing is correct here? I.e. that when disconnecting from a server, plugins are unlocked?
Overall, I think the `ShouldManageConfig` check should rather be somewhere else, for example in the `LockConfigurationEntries` method (not limited to that). `GetConfigFiles` is used at places that are timing indipendent

JotunnLib/Utils/ModCompatibility/ModCompatibility.cs
searica: All right, I'll look over the code and see what I can come up with. Currently none of the `ModuleVersionData` appear to be cached so it could take me a bit to figure out which instance to check against and ensure that it stays current. I appreciate the heads up that `LastServerVersion` can be set without mods though.

JotunnLib/Managers/SynchronizationManager.cs
searica: I did check the timings and seemed to work but I never checked how connecting to multiple servers with differing set ups within a single session worked and I could have missed something. I went for performing the check in `GetConfigFiles` since I thought it resulted in the least amount of duplicate code and effectively made it so Jotunn just didn't manage syncing a mod at all if Joutnn wasn't on the server.

Happy to change the check to be performing in the locking functions rather than in 'GetConfigFiles' though as it will probably prevent future issues and be an easier design to maintain and modify. 

I've since realized that I probably need to modify the `ShouldManageConfig` method to actually check if that specific mod is installed on the server though as simply having Jotunn installed on the server because some other mod needs it doesn't mean that users want to apply syncing to a mod that is only installed client-side.

searica:
I've reworked things to avoid locking config entries if a plugin that has `SynchronizationModeAttribute` set to `IfOnServer` is not installed on the server rather than just if Jotunn is not installed on the server.

To that end, I've added a method that explicitly checks if a plugin is installed on the server based on the plugin GUID. The check for whether Jotunn is installed now uses the method and simply calls it with the GUID for Jotunn. Implementing that method required adding the plugin GUID to `ModModule` to check against, as relying on the BepInPlugin.Name seemed like it could have unintended behaviour sometimes. This seemed like the most straightforward way to get a list of GUIDs of what is installed on the server but I'd appreciate any feedback on the approach.

`ShouldManageConfig` was reworked slightly to now check if the plugin is on the server rather than if Jotunn is on the server and the checks were moved from `GetConfigFiles` to only be in `LockConfigurationEntries` and `SetToDefaultConfigEntries` as those are the steps to be avoided is syncing is only meant to occur when the plugin is on both the server and client. There is no check in `UnlockConfigurationEntries` as unlocking configs that are already unlocked should have no effect and that avoids any possible timing issues with unlocking. Similarly I did not had a 'ShouldManageConfig' check in 'ResetAdminConfigs' as the boxed and local values should be the same if the config was never locked, set to default, or synced with the server values so it should have no effect and this also prevents possible timing issues.

I still need to do some more testing to make sure I haven't introduced any new issues but I wanted to push the changes to the PR so there was more opportunity to review them and provide feedback on the general approach if so desired.

JotunnLib/Utils/ModCompatibility/ModModule.cs
```diff
@@ -19,6 +21,7 @@ public ModModule(string name, System.Version version, CompatibilityLevel compati
 
         public ModModule(ZPackage pkg)
         {
+            guid = pkg.ReadString();
```
MSchmoecker: this will cause an error when an old Jotunn version connects to this version or vice versa (because ZPackage is very primitive and just reads data from the stream). Unfortunately, ModModule has no int version field for the data layout itself.

I'm rather reluctant to introduce such an issue (especially since it breaks the disconnect UI), after a Jotunn update the connections with mismatching versions is a lot higher. I think the only non-breaking way would be to add a new RPC first, with the changed data layout. If the new RPC is received, the second (old one) can be ignored. And connection is refused either way because the version missmatches

JotunnLib/Utils/ModCompatibility/ModCompatibility.cs
```diff
@@ -30,6 +32,34 @@ internal static void Init()
...
+            {
+                if (LastServerVersion == null)
+                {
+                    return false;
+                }
+                // maybe the HashSet of Module GUIDs should be cached along with LastServerVersion
+                // and reset along with it to, just to avoid construction of the HashSet
+                var serverData = new ModuleVersionData(LastServerVersion);
+                var moduleGUIDs = new HashSet<string>(serverData.Modules.Select(mod => mod.guid).ToList());
+                return moduleGUIDs.Contains(modGUID);
```
MSchmoecker: `IsModuleOnServer` is definitively called way too often to include such an expensive operation. This has to be cached, probably by always deserializing LastServerVersion instead of keeping the ZPackage

JotunnLib/Utils/ModCompatibility/ModModule.cs
searica: Oh, thanks for letting me know about this. I haven't dealt with RPCs much before and haven't run into issues with different versions of the same mod connecting before so I didn't think of that at all. I'm not entirely sure what you mean by adding a new RPC but I'll take a look when I have time over the next week and see if I can figure it out.

JotunnLib/Utils/ModCompatibility/ModCompatibility.cs
searica: I've just pushed some changes that implement and use a `ServerVersionData` class to hold the deserialized data and a cached copy of the GUID HashSet. I realize I may have to rollback the changes depending on how the RPC stuff pans out though as the `ServerVersionData` class relies on the new data layout of `ModModule` to get the GUIDs.

searica:
I've made some changes so that both the legacy and newly formatted data is serialized with the new format appended at the end for backwards compatibility. 

I think this should address most of the issues that were unresolved but it has introduced a new question, which is how to handle receiving data from a newer version of Jotunn that uses a `ModModule` data layout that is unknown. At the moment I've left it so all `ModModule` fields except for `dataLayoutVersion` are left as `null` and added a method to if a `ModModule` instance is formatted in one of the layouts supported by that version of Jotunn. Currently, I think that some kind of check needs to be added in `ModCompatibility` that triggers a disconnect when an unsupported data layout is received. 

I'm also not sure the best way to implement using the mod `GUID` rather than mod `name` for performing version checks such that it is backwards compatible but I am unsure if that should be part of this PR or not.

searica:
I've now added handling of when an unsupported data layout is received. Data layout is also considered when checking versions and a subsequent error message is provided in the disconnect window if there are unsupported data layout versions.

MSchmoecker (approved):

PR #446, opened 2025-02-16
Cache Local AdminOnly Configs Just Before Client Connects

searica:
Goal: 
Cache local configs as late as possible before connecting to a server and syncing config settings to avoid server settings overwriting local configs that are bound after the start scene.

Technical Details:
Replaced the `SceneManager_sceneLoaded` method that handled init/reset of admin with configs with `SetAdminConfigs_OnZNetStart`. Removed the event hook on scene change and added a patch to the internal private Patch class of `SynchronizationManager` to trigger a call to `SetAdminConfigs_OnZNetStart`.

### DIFF
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -178,10 +178,6 @@ namespace Jotunn.Managers
             [HarmonyPatch(typeof(ZNet), nameof(ZNet.Awake)), HarmonyPostfix]
             private static void ZNet_Awake(ZNet __instance) => Instance.ZNet_Awake(__instance);
 
-            // Hook ZNet.Start for handling lock/unlock of admin configs
-            [HarmonyPatch(typeof(ZNet), nameof(ZNet.Start)), HarmonyPrefix]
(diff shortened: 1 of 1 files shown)


MSchmoecker:
Sorry for the late response, been super busy lately.

ZNet.Start does't work for the `SceneManager.GetActiveScene().name == "start"` case, because the main menu does not start a ZNet. Thus, AdminOnly configs are locked in the main menu. This can be very confusing for players. I think FejdStartup is a good place to unlock entries.

The other case looks good and I don't see any potential timing issues. It actually fits great because we access ZNet.instance.

MSchmoecker:
I've moved the two cases into separate methods and removed the SceneManager check. ResetAdminState is now called before FejdStartup.Awake

Let me know if this still works for your case.

searica:
No worries, I totally get being busy. Good to know about the ZNet.Start not working. I must have gotten it mixed up with ZNetScene with regards to whether the main menu had an instance of it. 

I had considered splitting the lock and unlocking into two separate patches for timing reasons but thought it would be nice if they could still be bundled as one patch. The main issue for me was that caching happened too early so I think the approach you've proposed should work well.

PR #447, opened 2025-02-16
Add a ConfigFileWatcher Class

searica:
Goal:
Reduce unneeded boilerplate code that is copied between mods and provide a re-usable ConfigFile FileWatcher with events for mod authors.

Details:
Adds a new extension method `DisableSaveOnConfigSet()` to `ConfigFile` to disable save on config set while returning the value prior to calling the method so it can be cached and restored later.

Adds a new `ConfigFileWatcher` class to create a ConfigFile FileWatcher with events on a per config file basis.

### DIFF
--- a/JotunnLib/Documentation/tutorials/config.md
+++ b/JotunnLib/Documentation/tutorials/config.md
@@ -79,10 +79,10 @@ private void CreateConfigValues()
 
     // Add server config which gets pushed to all clients connecting and can only be edited by admins
     // In local/single player games the player is always considered the admin
-    Config.BindConfig("Server config", "StringValue1", "StringValue", "Server side string", true);
-    Config.BindConfig("Server config", "FloatValue1", 750f, "Server side float", true, floatRange);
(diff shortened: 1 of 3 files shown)


MSchmoecker:
I've made a few adjustments, most importantly make the `ConfigFileWatcher` public so mods can access it.

If you write some documentation for this feature, this is good to merge.

searica:
Changes all make sense to me, I also just learned about using `_` as a digit separator so thanks for both that and catching that `ConfigFileWatcher` wasn't public. 

Where and how would you like the documentation to be written?

MSchmoecker:
The documentation is located in the `JotunnLib/Documentation` folder, this fits in the `config.md`. A short usage example and explanation is enough, it should just be written down.

The markdown files are rendered in the online docs, if you are interested you can run them locally with docfx. To install docfx, run `dotnet tool install -g docfx --version 2.70.0` and then `docfx --serve` in the JotunnLib folder to open a local web server.

searica:
Thanks for pointing me to that. I have added some documentation in the config.md file for `ConfigFileWatcher` and I also added some for the `BindConfig` and `BindConfigInOrder` extension methods while I was at it.

PR #451, opened 2025-03-08
Fix: Config Sync Resetting for Local Worlds

searica:
Currently, local configs are cached whenever a player joins a server (including a solo world) and it resets the configs to the cached values when the player reaches the start menu. This means that when playing in a solo world, any config changes made while in the world are reset to the values they were during the main menu. 

I've addressed this by changing the hook for resetting the configs to the cached local values to trigger on `ZNet.OnDestroy` so that the reset logic can skip resetting to cached values if `ZNet.instance.IsServer() == true`. This should prevent configs being overwritten for solo play. I think it should also work to skip caching local values when joining a server if  `ZNet.instance.IsServer() == true` and I have left a comment to that effect.

The method `CacheConfigurationValues` has also been moved into `InitAdminState` as `ResetAdminState` now only runs for the first time when exiting a server rather than when entering the main menu.

While working on that I noticed a few likely unintended behaviours for config syncing as well. Config reloaded events were being subscribed to every single time a player reached the main menu so I have added a method to unsubscribe from the event and trigger the subscribing on `ZNet.Start` and the unsubscribing on `ZNet.OnDestroy`.

Similarly, the patches to `ConfigEntryBase` were being applied every time the main menu was reached so I have moved those patches into the `Patches` class so they are only applied once when `SynchronizationManager` first initializes.

### DIFF
--- a/JotunnLib/Managers/SynchronizationManager.cs
+++ b/JotunnLib/Managers/SynchronizationManager.cs
@@ -213,7 +213,7 @@ namespace Jotunn.Managers
             ///     Return the cached local value of a bep config thats locked
             /// </summary>
             [HarmonyPatch(typeof(ConfigEntryBase), nameof(ConfigEntryBase.GetSerializedValue)), HarmonyPrefix]
-            private bool GetCachedValueForSyncedConfigs(ConfigEntryBase __instance, ref string __result)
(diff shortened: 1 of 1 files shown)


MSchmoecker:
Thanks a lot, the code is sound and every case I've tested works exactly as expected.

PR #455, opened 2025-05-04
Add compatibility with SoftReferenceableAssets for customLocations

jneb802:
The goal of this PR is to change where a Location's prefab is stored. 

The current implementation of Jotunn.ZoneManager stores all CustomLocation prefabs in a LocationContainer GameObject, which is held in the DontDestroyOnLoaded scene. This approach causes CustomLocations prefabs to always be held in the client's memory and therefore leads to very high RAM usage when large numbers of CustomLocations are added. This PR transitions CustomLocation prefabs to use the [new vanilla SoftReferenceableAsset system](https://www.valheimgame.com/support/modding-faq-for-the-asset-bundle-update-0-217-40/), which allows prefabs to be loaded from AssetBundles on disk only when they are needed. 

This PR solves the above problem by making changes to three files in Jotunn. Mod developers will still need to follow additional steps to make their assetBundles available to the SoftReferenceableAssets system. The steps will be outlined in a future documentation page.

**CustomLocation.cs** 
- A new bool softReference field. This field is used in ZoneManager Harmony patches to check if mocks should be resolved for the customLocation.
- A new CustomLocation constructor which gets the CustomLocations SoftRef prefab.

**PrefabManager.cs**
- Add a new check if prefab name is has "JVLmock_". SoftReference mocks are resolved at a later point that before therefore many of the prefabs with the "JVLmock_" prefix were being automatically added to PrefabManager.

**ZoneManager.cs**
- Add a Harmony Prefix to ZoneSystem.SpawnLocation(). This prefix is where we now resolve mocks for softReference locations.
- Add a Harmony PostFix to ZoneSystemSpawnLocation() to add an additional Release(). We added an additional Load() call in our prefix so we must also include an additional Release().
- Add checks if customLocation use softReference to skip existing logic
- Add checks if prefab starts with "JVLmock_" to skip adding to PrefabManager

### DIFF
--- a/JotunnLib/Entities/CustomLocation.cs
+++ b/JotunnLib/Entities/CustomLocation.cs
@@ -4,6 +4,7 @@ using Jotunn.Configs;
 using Jotunn.Managers;
 using SoftReferenceableAssets;
 using UnityEngine;
+using Object = UnityEngine.Object;
 
 namespace Jotunn.Entities
 {
@@ -122,8 +123,9 @@ namespace Jotunn.Entities
                 Logger.LogError($"SoftReference invalid for prefab: {softReferencePrefab.Name}");
                 return;
             }
-            
-            AssetManager.Instance.ResolveMocksOnLoad(softReferencePrefab.m_assetID, ZoneManager.Instance.LocationContainer.transform);
+
+            var parent = ZoneManager.Instance.LocationContainer.transform;
+            AssetManager.Instance.ResolveMocksOnLoad(softReferencePrefab.m_assetID, parent, OnLocationResolve);
             Name = softReferencePrefab.Name;
             ZoneLocation = locationConfig.GetZoneLocation();
             ZoneLocation.m_prefab = softReferencePrefab;
@@ -131,7 +133,15 @@ namespace Jotunn.Entities
             FixReference = fixReference;
             SoftReference = true;
         }
-        
+
+        private void OnLocationResolve(Object asset)
+        {
+            if (asset is GameObject gameObject && gameObject.TryGetComponent<ZoneSystem.ZoneLocation>(out var zoneLocation))
+            {
+                ZoneManager.Instance.PrepareLocation(zoneLocation, SourceMod);
+            }
+        }
+
         /// <summary>
         ///     Helper method to determine if a location prefab with a given name is a custom location created with Jötunn.
         /// </summary>
--- a/JotunnLib/Entities/CustomLocation.cs
+++ b/JotunnLib/Entities/CustomLocation.cs
@@ -103,7 +103,7 @@ namespace Jotunn.Entities
             }
 
             ZoneLocation = locationConfig.GetZoneLocation();
-            ZoneLocation.m_prefab = AssetManager.Instance.InjectLoadedAsset(exteriorPrefab);
+            ZoneLocation.m_prefab = new SoftReference<GameObject>(AssetManager.Instance.AddAsset(exteriorPrefab));
             ZoneLocation.m_prefabName = exteriorPrefab.name;
 
             FixReference = fixReference;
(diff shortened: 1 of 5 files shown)


JotunnLib/Entities/CustomLocation.cs
```diff
@@ -103,6 +109,30 @@ public CustomLocation(GameObject exteriorPrefab, GameObject interiorPrefab, bool
             FixReference = fixReference;
         }
         
+        /// <summary>
+        ///     Custom location from a prefab with a <see cref="LocationConfig"/> attached. Using SoftReference system.
+        /// </summary>
+        /// <param name="prefabName">The exterior prefab for this custom location.</param>
+        /// <param name="fixReference">If true references for <see cref="Entities.Mock{T}"/> objects get resolved at runtime by Jötunn.</param>
+        /// <param name="locationConfig">The <see cref="LocationConfig"/> for this custom location.</param>
+        public CustomLocation(string prefabName, bool fixReference, LocationConfig locationConfig) : base(Assembly.GetCallingAssembly())
```
MSchmoecker: It should be clearer this is using the SoftReferenceSystem. Maybe we should pass `SoftReference<GameObject>` as the first parameter?

JotunnLib/Managers/ZoneManager.cs
```diff
@@ -110,6 +110,36 @@ private static void ZoneSystem_SetupLocations(ZoneSystem __instance)
...
+                if (CustomLocation.IsCustomLocation(location.m_prefab.Name))
+                {
+                    CustomLocation customLocation = Instance.GetCustomLocation(location.m_prefab.Name);
+                    if (customLocation.FixReference && customLocation.SoftReference)
+                    {
+                        location.m_prefab.Load();
+                        GameObject mockLocationContainer = Object.Instantiate(location.m_prefab.Asset, Instance.LocationContainer.transform);
+                        mockLocationContainer.FixReferences(true);
+                        location.m_prefab.m_loadedAsset = mockLocationContainer;
+                        Object.Destroy(mockLocationContainer);
```
MSchmoecker: The destroy is a somewhat obscure way to unload the location and may lead to an inconsistent state.
If something else holds a reference to the location, the internal reference counter will be greater than 1, but `location.m_prefab.m_loadedAsset` is destroyed at the next frame. So the SoftReferenceSystem will not try to load the asset again and instead NREs will be thrown at unrelated places.

JotunnLib/Entities/CustomLocation.cs
jneb802: Ok sure I made that change

JotunnLib/Managers/ZoneManager.cs
jneb802: Why is `location.m_prefab.m_loadedAsset` destroyed on the next frame?

My impression is that this Destroy has nothing to do with the SoftReference Asset. Rather, it is destroying extra container that we're using to resolve the locations mocks.

MSchmoecker: Yes, but the code is setting `m_loadedAsset = mocklocationContainer` which is then marked to be destroyed at the next frame. Unless I'm missing something

jneb802: Ok I understand now. My impression is we need to remove this mockLocationContainer from the Dont'DestroyOnLoad scene somehow. If we don't then it creates limitless mockLocationContainers. Do you have a suggestion on another way we could remove it or maybe another way to resolve the mocks?

MSchmoecker: We could move the Destroy to the unload phase, if we are certain the underling asset isn't loaded anymore.

I haven't tested this code path but it could be for example:
```cs
GameObject loadedAsset = location.m_prefab.m_loadedAsset;
location.m_prefab.Release();
if (!location.m_prefab.IsLoaded && loadedAsset)
{
    Object.Destroy(loadedAsset);
}
```

We also have to make sure to prevent creating orphan assets in the load phase, if `location.m_prefab` was already loaded. Either by checking if location.m_prefab is not already loaded or by creating a List/HashSet of already resolved locations.

PR #456, opened 2025-05-11
Fix TryTranslate throwing warnings if the input is an already localized string

MSchmoecker:


OrianaVenture:
This seems to fix our problem. I am still concerned that failure of this method would potentially cause log spamming. Is there a strong reason to needing the warning log? If yes can that be converted to debug log level?

MSchmoecker:
The warning is good to tell when an incorrect keys is used. Putting this as a debug log doesn't seem useful.

However, I can change it to only log it once for any given token. That logs the warning but prevents spamming

MSchmoecker:
@OrianaVenture the log spam is now prevented by logging the warning once per token. Is there anything else or is this good to merge?

OrianaVenture:
This should work great thank you!

PR #457, opened 2025-05-12
Fix: Allow mocking to fail on dedicated server builds.

OrianaVenture:
When mocking fails on a dedicated server it should still add the mocked entity to the appropriate systems. This will fix locations not existing/being placed on dedicated server when mocking fails (example: Mining Caves version 0.2.0).


MSchmoecker:
I'm aware the mocks are currently only broken on a dedicated server but the mocking behavior should rather not be influenced where it's loaded. The dedicated check is rather suited for networking cases.

As discussed, it's a valid decision to change the design to only log a warning and return null if a mock cannot be resolved. I think that's a better solution. In many cases unresolved mocks are irrelevant anyway, otherwise the calling mod is responsible for downstream errors.

Of course we miss some context which is currently logged at the top level. We can either pass that down or use a static variable to set context before resolving the mocks. At least the top-level prefab name has to be included in the warning.

OrianaVenture:
Having a way to expose a list of all the failed mocks will make debugging faster. Currently it will only show one so this will be a good change to make.

By making the log level warning level my mining caves mod would log about 500 warnings about mock failures on dedicated servers. I feel this will cause a lot of questions and user support for information that users do not need to necessarily see. If we can leave the specifics as debug logs and log one general warning about mock failures that would solve my problem.

I will rework this with the above change and remove the dedicated check.

MSchmoecker:
Yeah, listing all failed mocks is a positive side effect.

The current situation on dedicated servers is (hopefully) only temporary. It should not make logging more obscure, the Debug level is not visible by default.
However, I'm open to disable the failed mocks warning per mod (or similar solutions). E.g. the mod can call something like `MockManager.Instance.LogFailedMockWarnings(false)`, or in your case you are free to call `MockManager.Instance.LogFailedMockWarnings(!ZNet.instance.IsDedicated())` on your side. We have `BepInExUtils.GetSourceModMetadata()` to get the calling BepInEx mod and don't need to pass it around.

OrianaVenture:
If you want to implement disabling the warnings I could be open to that. It seems like an overengineered solution when the debug level is available and we could be utilizing it throughout the library as a solution. Given Valheim updates could cause log spam like this at any time it makes more sense to me that advanced information should just be hidden in debug logs that any author could turn on and off as needed rather than having to recompile the mod to change the log levels. On a game update it would be much faster to just boot up with my normal debugging environment to see what's changed, not compile a debug version of my mod for advanced testing

Teaching other authors how to turn on debug log levels is much easier than adding library features in my opinion too. I feel a message like "500 Mocks could not be resolved. To view more information enable Debug log levels in your bepinex config." Or even having a configuration in Jotunn to toggle rather than bepinex debugging would work.

MSchmoecker:
That's a fair point, disabling the warning per mod is indeed overengineered.

My point is that failed mocks are warnings and not debug information. They can lead to downstream errors and having a related warning before an error is very valuable. The most simple solution is to just log them as warnings.

But of course I understand that unavoidable log spam is annoying. However, that is an exceptional case duo to the broken dedicated server entry point. In the normal case, we don't have log spam with failed mocks. Having precise warning/error logs helps with troubleshooting for both players and devs.

As a compromise, could we collect the failed mocks and display an aggregated message if there are more than like 5-10 of them, otherwise displaying them individually?

OrianaVenture:
I could be comfortable with truncating the logs like that. Let me implement this and send for review.

OrianaVenture:
Example output:

```log
[Debug  :Jotunn.Managers.ZoneManager] Adding custom location VV_CopperTinCave in BlackForest, All
[Warning:Jotunn.Managers.MockSystem.MockResolveFailure] 260 mocks could not be resolved. Details may be truncated.
[Info   : Unity Log] 05/14/2025 14:08:39: Console: [Warning:Jotunn.Managers.MockSystem.MockResolveFailure] 260 mocks could not be resolved. Details may be truncated.

[Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.
[Info   : Unity Log] 05/14/2025 14:08:39: Console: [Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.

[Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.
[Info   : Unity Log] 05/14/2025 14:08:39: Console: [Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.

[Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.
[Info   : Unity Log] 05/14/2025 14:08:39: Console: [Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.

[Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.
[Info   : Unity Log] 05/14/2025 14:08:39: Console: [Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'exirt_portal' could not be resolved. GameObject with name 'exirt_portal' was not found.

[Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'Curvedrock_mat' could not be resolved. GameObject with name 'Curvedrock_mat' was not found.
[Info   : Unity Log] 05/14/2025 14:08:39: Console: [Warning:Jotunn.Managers.MockSystem.MockResolveFailure] Mock Material 'Curvedrock_mat' could not be resolved. GameObject with name 'Curvedrock_mat' was not found.
```

MSchmoecker:
1. Can you move the MockResolveFailure.ClearMockResolveFailures/Print into PrefabExtension.FixReferences(this object objectToFix) and PrefabExtension.FixReferences(this GameObject gameObject, bool recursive), the two main entry points of fixing references? So we don't need to clear and print everything manually. Plus when a mod calls FixReferences the logs are also correct. Of course we loose some fine control but that should be alright.
2. MockResolveFailure and MockResolveException are way too similar, please remove either one

OrianaVenture:
Same test results after the refactor. Is there any other cleanup needed for this?

MSchmoecker:
I have some minor code changes that are not worth discussing. Since you used your dev branch for the PR, Github seems to forbid me pushing the commit here. Even tho "Maintainers are allowed to edit this pull request" is allowed. Will just push it to dev after the PR is merged.

Will do some testing for some edge cases and then merge everything at the weekend.

PR #460, opened 2025-07-12
Add new GenerateAssetID overload

jneb802:
Adds a new GenerateAssetID overload to accept a string param instead of Object.


MSchmoecker:
For API consistency, the method should not be static but be part of the singleton instance

jneb802:
Ok I made it not static

PR #463, opened 2025-10-22
Vulkan compile target

MidnightsFX:
Change unity project APIs to not be auto and set Vulkan as a target API.
Upgrade unity to 6000.0.58f2


MSchmoecker:
Thanks!

PR #467, opened 2026-01-15
feat: Add macOS development support

jneb802:
- Add VALHEIM_MANAGED variable to Paths.props with OS-agnostic defaults
- Replace hardcoded Windows paths in JotunnLibRefsCorlib.props with variable references
- Use forward slashes for cross-platform compatibility
- Support Environment.props import for local path overrides
- Remove VALHEIM_MANAGED override from JotunnLib.csproj

### DIFF
--- a/JotunnLib/BuildProps/Paths.props
+++ b/JotunnLib/BuildProps/Paths.props
@@ -23,7 +23,9 @@
   <PropertyGroup Condition="$(BEPINEX_PATH)=='' AND $(VALHEIM_INSTALL) != ''">
     <BEPINEX_PATH>$(VALHEIM_INSTALL)\BepInEx</BEPINEX_PATH>
   </PropertyGroup>
+  <!-- Set VALHEIM_MANAGED with Exists check for both capitalizations -->
   <PropertyGroup Condition="$(VALHEIM_MANAGED)=='' AND $(VALHEIM_INSTALL) != ''">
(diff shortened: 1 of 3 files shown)


JotunnLib/BuildProps/Paths.props
```diff
@@ -22,4 +23,7 @@
   <PropertyGroup Condition="$(BEPINEX_PATH)=='' AND $(VALHEIM_INSTALL) != ''">
     <BEPINEX_PATH>$(VALHEIM_INSTALL)\BepInEx</BEPINEX_PATH>
   </PropertyGroup>
+  <PropertyGroup Condition="$(VALHEIM_MANAGED)=='' AND $(VALHEIM_INSTALL) != ''">
+    <VALHEIM_MANAGED>$(VALHEIM_INSTALL)/valheim_Data/Managed</VALHEIM_MANAGED>
```
MSchmoecker: It should be `Valheim_Data/Managed` here. Maybe both paths can be added with an `Exists` condition, in case someone has an old folder or Valheim changes it back

jneb802: @MSchmoecker Good point, addressed that in:

https://github.com/Valheim-Modding/Jotunn/pull/467/commits/1e364126618183d34d16dbef424ddb4b8522064b

PR #472, opened 2026-04-13
fix: sync ZoneLocation radii/ClearArea from prefab's Location component

jneb802:
## The Issue

`LocationConfig.ExteriorRadius`, `InteriorRadius`, and `ClearArea` each exist in two independent places:

1. The `Location` MonoBehaviour on the prefab.
2. The `ZoneSystem.ZoneLocation` registration record.

`CustomLocation` sourced `ZoneLocation.m_exteriorRadius` exclusively from `LocationConfig.ExteriorRadius` (defaulting to `10f`). The prefab's `Location` component was preserved when present but never read into the `ZoneLocation`, so the two values could silently diverge.

## Where It Surfaced

The mismatch was caught via [VentureValheim/LocationReset](https://github.com/OrianaVenture/VentureValheim/tree/master/LocationReset), which resets the contents of a location to their originally generated state. Its `LocationReset.LocationPosition` constructor reads the reset boundary from `ZoneSystem.ZoneLocation`:

```csharp
public LocationPosition(LocationProxy loc, ZoneSystem.ZoneLocation zone, Location location)
{
    ...
    GroundDistance = zone.m_exteriorRadius;   // ZoneLocation, NOT the Location component
    SkyDistance    = zone.m_interiorRadius;
    ...
}
```

LocationReset calls `TryResetAfterLoadPrefab`, with `zone` fetched via `ZoneSystem.instance.GetLocation(hash)`. For [More World Locations AIO](https://thunderstore.io/c/valheim/p/warpalicious/More_World_Locations_AIO/) locations, `zone.m_exteriorRadius` carried the `LocationConfig` value while the radius set in UnityEditor was larger. Therefore, LocationReset only reset a subset of the location's actual footprint and left objects beyond the config radius unreset.

### `LocationConfig`

The three shared fields become nullable so Jötunn can distinguish "caller explicitly set this" from "caller left this unset." When a field is `null`, `CustomLocation` treats the prefab's `Location` component as the source of truth; when it has a value, that value wins.

- `GetZoneLocation()` falls back to the prior defaults when unset, preserving behavior for prefabs without a `Location` component. Existing callers that assign a value compile unchanged via implicit conversion from `float` / `bool` to the nullable types.

### `CustomLocation`

Synchronizes the three shared fields onto the `ZoneLocation` from the prefab's `Location` component whenever the corresponding config field is `null`. **Config wins when set; otherwise the component's value flows through.**

- **Non-SoftReference constructor:** `SyncZoneLocationFromComponent()` runs immediately after `GetZoneLocation()`.
- **SoftReference constructor:** `SyncZoneLocationFromComponent()` runs inside `OnLocationResolve(GameObject)` once the asset loads, mutating the already-registered `ZoneLocation` in place. Since `ZoneLocation` is a reference type, `ZoneSystem.m_locations` sees the update.

### DIFF
--- a/JotunnLib/Entities/CustomLocation.cs
+++ b/JotunnLib/Entities/CustomLocation.cs
@@ -95,11 +95,11 @@ namespace Jotunn.Entities
             else
             {
                 Location = exteriorPrefab.AddComponent<Location>();
-                if (locationConfig.ClearArea.HasValue) Location.m_clearArea = locationConfig.ClearArea.Value;
-                if (locationConfig.ExteriorRadius.HasValue) Location.m_exteriorRadius = locationConfig.ExteriorRadius.Value;
+                if (locationConfig.HasClearArea) Location.m_clearArea = locationConfig.ClearArea;
+                if (locationConfig.HasExteriorRadius) Location.m_exteriorRadius = locationConfig.ExteriorRadius;
                 Location.m_interiorPrefab = interiorPrefab;
                 Location.m_hasInterior = locationConfig.HasInterior;
-                if (locationConfig.InteriorRadius.HasValue) Location.m_interiorRadius = locationConfig.InteriorRadius.Value;
+                if (locationConfig.HasInteriorRadius) Location.m_interiorRadius = locationConfig.InteriorRadius;
                 Location.m_interiorEnvironment = locationConfig.InteriorEnvironment;
             }
 
@@ -159,9 +159,9 @@ namespace Jotunn.Entities
                 return;
             }
 
-            if (!locationConfig.ExteriorRadius.HasValue) ZoneLocation.m_exteriorRadius = location.m_exteriorRadius;
-            if (!locationConfig.InteriorRadius.HasValue) ZoneLocation.m_interiorRadius = location.m_interiorRadius;
-            if (!locationConfig.ClearArea.HasValue) ZoneLocation.m_clearArea = location.m_clearArea;
+            if (!locationConfig.HasExteriorRadius) ZoneLocation.m_exteriorRadius = location.m_exteriorRadius;
+            if (!locationConfig.HasInteriorRadius) ZoneLocation.m_interiorRadius = location.m_interiorRadius;
+            if (!locationConfig.HasClearArea) ZoneLocation.m_clearArea = location.m_clearArea;
         }
 
         /// <summary>
--- a/JotunnLib/Entities/CustomLocation.cs
+++ b/JotunnLib/Entities/CustomLocation.cs
@@ -15,6 +15,8 @@ namespace Jotunn.Entities
     /// </summary>
     public class CustomLocation : CustomEntity
     {
+        private readonly LocationConfig _locationConfig;
+
         /// <summary>
         ///     The exterior prefab for this custom location.
         /// </summary>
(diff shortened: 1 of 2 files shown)


OrianaVenture:
Other custom managers could also benefit from a similar feature! I would argue all fields should be set this way, where if they do not exist it pulls the value from the prefab rather than setting a default value. It will help speed up development so we do not need to specify values in both Unity and in code. Specifying in code can be used for quick testing of changing config values or updates/overrides after the asset bundle has been created.

MSchmoecker:
Unfortunately, a change from `float` to `float?` breaks backwards compatibility because C# threats them as different types.

E.g. arbitrarily with RtDMonsters, but any other mod that access one of the changed types:
```
[Info   :   BepInEx] Loading [Jotunn 2.29.0]
[Info   :Jotunn.Main] Initializing ModCompatibility
[Info   :Jotunn.Main] Initializing SynchronizationManager
[Info   :Jotunn.Main] Initializing NetworkManager
[Info   :   BepInEx] Loading [RtDMonsters 2.4.13]
[Info   :Jotunn.Main] Initializing ItemManager
[Info   :Jotunn.Main] Initializing PrefabManager
[Info   :Jotunn.Main] Initializing MockManager
[Info   :Jotunn.Main] Initializing LocalizationManager
[Info   :Jotunn.Main] Initializing AssetManager
[Info   :Jotunn.Main] Initializing CreatureManager
[Error  : Unity Log] MissingMethodException: Method not found: void Jotunn.Configs.LocationConfig.set_ExteriorRadius(single)
Stack trace:
RtDMonsters.RtDMonsters.Awake () (at <fc754e8b679f45f098bf3c74bdd97f8b>:0)
UnityEngine.GameObject:AddComponent(Type)
BepInEx.Bootstrap.Chainloader:Start() (at C:/Users/crypt/RiderProjects/BepInEx/BepInEx/Bootstrap/Chainloader.cs:434)
UnityEngine.GameObject:.cctor()
PlatformInitializer:EarlyInitialize()
```
It's a MissingMethodException because C# converts getter/setter to methods internally.

jneb802:
@MSchmoecker I made a change to address backward compatibility and tested with [RtDMonsters](https://thunderstore.io/c/valheim/p/ReadyToDie/RtDMonsters/). I did not get the MissingMethodException error you shared above.

JotunnLib/Entities/CustomLocation.cs
```diff
@@ -95,18 +95,20 @@ public CustomLocation(GameObject exteriorPrefab, GameObject interiorPrefab, bool
...
             {
                 Location = exteriorPrefab.AddComponent<Location>();
-                Location.m_clearArea = locationConfig.ClearArea;
-                Location.m_exteriorRadius = locationConfig.ExteriorRadius;
+                if (locationConfig.HasClearArea) Location.m_clearArea = locationConfig.ClearArea;
+                if (locationConfig.HasExteriorRadius) Location.m_exteriorRadius = locationConfig.ExteriorRadius;
                 Location.m_interiorPrefab = interiorPrefab;
                 Location.m_hasInterior = locationConfig.HasInterior;
-                Location.m_interiorRadius = locationConfig.InteriorRadius;
+                if (locationConfig.HasInteriorRadius) Location.m_interiorRadius = locationConfig.InteriorRadius;
```
MSchmoecker: Shouldn't this stay as it was before and not check the HasClearArea etc flags?
The `Location` component is freshly created and thus uses initializes with default values, instead of the fallback values from locationConfig

jneb802: @MSchmoecker Yes, that's a good point. I made a change to that

MSchmoecker (approved):
Thanks, this is good to merge now!

PR #474, opened 2026-05-04
Fix for Minimap texture copy warning

MidnightsFX:
Fix texture mismatch when writing multiple layers, avoid format incompatibilities.

This fixes the discrepancy created when copying textures currently (CPU vs GPU do not match). 

```
[Warning: Unity Log] Graphics.CopyTexture called with mismatching CPU texture memory sizes, perhaps formats are incompatible? Pixel data was successfully copied on the GPU, but not copied on the CPU. (srcsize=8388608 dstsize=16777216)
```

Which is caused by using the minimap overlay system. Currently creating an overlay causes this warning each time the texture is written.

While technically harmless unless another Copy(), or GetPixels() is called, if either is invoked it will use stale data from the CPU cache (which is outdated).

To fix this and address the formatting issues which cause it, this performs just a GPU -> GPU copy.

If you would like to test or reproduce, [StarLevelSystems](https://thunderstore.io/c/valheim/p/MidnightMods/StarLevelSystem/) includes an overlay by default which runs into this issue.

### DIFF
--- a/JotunnLib/Managers/MinimapManager.cs
+++ b/JotunnLib/Managers/MinimapManager.cs
@@ -333,7 +333,7 @@ namespace Jotunn.Managers
                     if (Overlays.Values.Any(x => !x.IgnoreFog && x.Dirty) ||
                         Drawings.Values.Any(x => x.FogEnabled && x.FogDirty))
                     {
-                        Graphics.CopyTexture(Minimap.instance.m_fogTexture, FogFilter);
+                        BlitCopy(Minimap.instance.m_fogTexture, FogFilter);
                     }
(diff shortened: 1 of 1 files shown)


MSchmoecker:
The underlying reason for the error seems to be because Valheim changed the TextureFormat of the fog texture. Copying the texture over via a `RenderTexture` and `Graphics.Blit` is more expensive than a `Graphics.CopyTexture` call. So the real fix is to adjust the textures to match the vanilla format again.

MidnightsFX:
Closing this to replace with just texture format changes as requested. https://github.com/Valheim-Modding/Jotunn/pull/476

PR #476, opened 2026-05-28
Minimap texture format fixes

MidnightsFX:
reference the existing minimap textures instead of hardcoding to prevent future texture change mismatches in minimap rendering.


MSchmoecker (approved):
That's a better solution, thanks!

PR #480, opened 2026-09-04
fix: resolve the Valheim data folder case-insensitively in the prebuild task

joeyparrish:
## Problem

`JotunnBuildTask` hardcodes the game data folder as `Valheim_Data`:

```csharp
internal const string ValheimData = "Valheim_Data";
```

On Linux the Steam install ships `valheim_Data` with a lowercase `v`, and `Directory.Exists` is case-sensitive there, so the prebuild task never finds the `Managed` folder and publicizing fails for every Linux mod build.

`JotunnLib/BuildProps/Paths.props` already probes both spellings when it sets `VALHEIM_MANAGED` (1e36412), so the props and the build task disagreed about which install layouts are supported. CI does not catch this because the workflows rename `valheim_server_Data` to `Valheim_Data` before building, which lands on the one spelling the task accepts.

The symptom is unhelpfully vague too. The task wrote its message to `Console.WriteLine` and returned false without logging an error, so MSBuild reported only:

```
error MSB4181: The "JotunnBuildTask" task returned false but did not log an error.
```

The real message (`<path> does not include Valheim_Data or valheim_server_Data`) scrolls by as plain build output, and it does not mention letter case at all.

## Changes

1. **Resolve the data folder case-insensitively.** Both places that resolved the `Managed` folder, `Execute()` and `AssemblyPublicizer.PublicizeDll`, now share a single `GetManagedFolders` helper. It enumerates the install folder and matches the data folder name case-insensitively, so `Valheim_Data`, `valheim_Data` and any other casing all resolve. Existing precedence is preserved: the client folder wins over `valheim_server_Data`.

2. **Log failures as MSBuild errors.** `Execute()` now uses `Log.LogError` so MSBuild surfaces the actual reason instead of MSB4181. This is a separate commit and can be dropped if you would rather keep the PR narrower.

`libraries/JotunnBuildTask.dll` is rebuilt in each commit, since the props point at the checked-in binary. It was built with the .NET 10 SDK targeting net462. Happy to drop those two blobs if you would prefer to rebuild it yourselves.

## Verification

All on Linux, driving the task from a trivial mod project with `<ExecutePrebuild>true</ExecutePrebuild>`.

Against a real Steam install (`valheim_Data`):

- Before the change: `error MSB4181`, no publicized assemblies produced.
- After: build succeeds and `valheim_Data/Managed/publicized_assemblies/` is populated with all 10 `*_publicized.dll` files, including `SoftReferenceableAssets_publicized.dll` and `gui_framework_publicized.dll`.
- Deleting `publicized_assemblies/` and rebuilding regenerates it from a clean state; a second build correctly skips republicizing via the hash check.
- `JotunnLib.csproj` itself now builds end to end on Linux and produces `Jotunn.dll`.

Against synthetic install layouts:

| Layout | Result |
| --- | --- |
| `Valheim_Data/Managed` | resolves (the CI layout) |
| `valheim_Data/Managed` | resolves (was broken) |
| `valheim_server_Data/Managed` | resolves, full publicize run verified |
| `VaLhEiM_dAtA/Managed` | resolves |
| both `valheim_Data` and `valheim_server_Data` | client folder wins, as before |
| neither | fails, now with the real message instead of MSB4181 |

I could not find an existing issue for this, so I did not link one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


joeyparrish:
Heads up on the CI failure, since it does not come from this diff. The job fails in `Prepare Valheim dependencies`, before `Build solution` ever runs, and this branch only touches `JotunnBuildTask/*.cs` and `libraries/JotunnBuildTask.dll`. It is not a transient flake and re-running the job will not clear it, so I dug into it.

### `steamcmd` now fails on any cold cache

```
ERROR! Failed to install app '896660' (Missing configuration)
mv: cannot stat '/home/runner/VHINSTALL/valheim_server_Data/': No such file or directory
```

The `mv` error is just the cascade from nothing having been installed.

I reproduced this locally with steamcmd 1788292693, the same build the runner reported, running the workflow's exact command against a fresh `HOME`:

1. First invocation: `Failed to install app '896660' (Missing configuration)`
2. Second invocation, identical command, same now-warm steamcmd state: `Success! App '896660' fully installed.`

So the first `app_update` against a fresh steamcmd fails, and it works once app info has been fetched. Runners are ephemeral, so every re-run starts cold and fails identically.

Fetching app info first fixes it. Verified from a clean `HOME`:

```yaml
steamcmd +login anonymous +app_info_update 1 +app_info_print 896660 +quit
steamcmd +force_install_dir ~/VHINSTALL +login anonymous +app_update 896660 validate +exit
```

The first line succeeds cold, and the second then downloads and installs normally. Repeating the `app_update` line instead works too.

All three workflows use the identical invocation (`pull-request.yml:47`, `merge.yml:56`, `tag-release.yml:57`), and `/repos/Valheim-Modding/Jotunn/actions/caches` currently reports `total_count: 0`, so the next merge to `dev` and the next tag release look likely to fail the same way, not just PRs.

### Why it surfaced now: the PR cache key never includes the version

`pull-request.yml:18`:

```yaml
- name: Get Valheim version id
  id: valheimversion
  run: echo "valheimversion=$(curl -s https://api.steamcmd.net/v1/info/896660 | jq -r ".data.\"896660\".depots.branches.public.buildid")"
```

That echoes to stdout and never appends to `$GITHUB_OUTPUT`, so `steps.valheimversion.outputs.valheimversion` is empty and the cache key collapses to the constant `--BepInExPack-5.4.2333`. The log confirms it:

```
Cache not found for input keys: --BepInExPack-5.4.2333
```

`merge.yml:27` and `tag-release.yml:21` both have the `>> $GITHUB_OUTPUT` redirect, so this one is specific to the PR workflow. Two consequences: the PR cache never invalidates when Valheim updates, so PR builds can run against an arbitrarily stale install, and while that constant key kept hitting, the steamcmd path never ran, which is what kept the problem above hidden until the cache aged out.

I plan to send a separate PR fixing both of these. I have kept this one limited to the publicizer fix.

joeyparrish:
PR #481 fixes the workflow, after which we can merge `dev` into this branch and run CI again.

joeyparrish:
Thank you for the quick turnaround on a new release!

PR #481, opened 2026-09-04
ci: fix Valheim dependency setup on a cold cache

joeyparrish:
On a fresh steamcmd, app_update fails immediately with "ERROR! Failed to install app '896660' (Missing configuration)". The following mv then fails because nothing was installed, and the job exits 1 before the solution is ever built. Adding +app_info_update 1 before +app_update fixes it. Runners are ephemeral, so every run and re-run starts from that cold state, and re-running a failed job does not help.

This was masked until now because the PR workflow echoed the Valheim version id to stdout instead of appending it to $GITHUB_OUTPUT, so steps.valheimversion.outputs.valheimversion was always empty and the cache key collapsed to the constant "--BepInExPack-5.4.2333". That constant key kept getting hit, so the steamcmd path never ran. It also meant the cache never invalidated when Valheim updated, so PR builds could run against an arbitrarily stale install. merge.yml and tag-release.yml already had the redirect.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -53,7 +53,9 @@ jobs:
         run: |
           wget -O bepinex.zip "https://thunderstore.io/package/download/denikson/BepInExPack_Valheim/5.4.2333/"
           unzip bepinex.zip -d ~/BepInExRaw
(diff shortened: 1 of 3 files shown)


MSchmoecker:
Thanks, the CI always failed randomly with this. Since it's hard to reproduce, this hopefully did the trick!

PR #483, opened 2026-09-09
Fix Valheim 1.0 compatibility

jneb802:
## Summary

- update Jotunn for Valheim 1.0 sector, terrain, equipment, network-version, and piece-table APIs
- update bundled test-mod code for the new map and equipment types
- prepare the actual ZoneLocation after a soft-reference location resolves
- reject invalid soft-reference locations and rooms without accessing an unavailable asset name

This pull request intentionally excludes the macOS publicizer and build-path changes. Those remain on the mac-dev branch for local development only.

## Validation

- Valheim 1.0.7, Steam build 25185596
- dotnet build JotunnLib.sln --no-restore -p:ExecutePrebuild=false: passed
- xUnit console: 28 passed, 0 failed
- Valheim 1.0 startup reproduced the invalid AssetID location exception before the fix

The updated soft-reference behavior has not yet been validated by spawning a custom location or generating a custom dungeon. The build retains existing netstandard reference and source warnings. No warning originates from these changes.

PR #484, opened 2026-09-11
fix: register custom TerrainOp prefabs in ObjectDB

MaikiOS:
## Problem

Valheim 1.0 serializes `TerrainOp.Settings` using the prefab's stable hash. On the receiving side, `TerrainOp.Settings.Deserialize` resolves that hash only through `ObjectDB.TryGetTerrainOp`. Custom prefabs registered by Jotunn are added to `ZNetScene`, but their `TerrainOp` components are not added to `ObjectDB.m_terrainOps` / `m_terrainOpsByHash`, so deserialization fails and the terrain operation is cancelled.

This affects custom terrain prefabs in multiplayer and after network serialization. The issue was reproduced with [AdvancedTerrainModifiersCompatible 1.4.3](https://thunderstore.io/c/valheim/p/Ostrix/AdvancedTerrainModifiersCompatible/) using `raise_v2_precise` and `paved_road_v2_square`.

## Fix

Register a prefab's `TerrainOp` in both ObjectDB collections from the existing `PrefabManager.RegisterToZNetScene` path. Registration occurs only when the exact prefab instance is the one stored in `ZNetScene`, preserving the existing duplicate/hash-collision behavior.

No new public API or dependency is introduced.

## Validation

- `JotunnLib` compile target passes against the Valheim 1.0.7 assemblies.
- Runtime-tested on Valheim 1.0.7 with both affected terrain prefabs: placement, save, and clean shutdown succeeded.
- The runtime log contained no `Failed to deserialize TerrainOp settings`, exceptions, errors, or prefab hash collisions.
- `git diff --check` passes.

The original investigation and proposed diff were discussed in [#483](https://github.com/Valheim-Modding/Jotunn/pull/483#issuecomment-5607892217), but the TerrainOp registration was not included in that release fix.

### DIFF
--- a/JotunnLib/Managers/PrefabManager.cs
+++ b/JotunnLib/Managers/PrefabManager.cs
@@ -381,7 +381,7 @@ namespace Jotunn.Managers
                 
                 int hash = name.GetStableHashCode();
 
-                if (znet.m_namedPrefabs.ContainsKey(hash))
+                if (znet.m_namedPrefabs.TryGetValue(hash, out GameObject registeredPrefab))
                 {
                     Logger.LogDebug($"Prefab {name} already in ZNetScene");
                 }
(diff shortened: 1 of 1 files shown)


MSchmoecker:
Thanks, I just changed the code a bit so we don't have to pass the registeredPrefab around. Otherwise is good.

PR #488, opened 2026-09-19
fix: don't throw in AssetManager transpiler when Dictionary.Add was already replaced

### LINKED ISSUE #485: AssetManager: GetAllAssetPathsMappedToAssetID transpiler throws (and kills AssetManager) when Dictionary.Add was already rewritten by another mod
`AssetManager.Patches.AssetBundleLoader_GetAllAssetPathsMappedToAssetID` calls `SetInstruction` without checking `IsValid` after `MatchForward`. When another mod has already rewritten the `Dictionary.Add` (STUWard 1.3.15 does), the exception kills `AssetManager`'s static constructor and every content mod that depends on it. Guarding with `ThrowIfInvalid`→warning, or `if (matcher.IsValid)`, would degrade gracefully instead.

Seen with STUWard 1.3.15 (see its `WardUiResources.AssetPathsPatch`) on Valheim 1.0.12 (n-40), BepInExPack 5.4.2350, Jotunn 2.30.0. Every Jotunn content mod (e.g. ValheimRAFT 4.3.1) then fails with `TypeInitializationException: The type initializer for 'Jotunn.Managers.AssetManager' threw an exception`.

Also reported on the STUWard side: sighsorry1029/STUWard#2

Hampus-Toft:
Fixes #485

**Disclosure:** I used Claude (Anthropic) to help investigate this and prepare the change. I ran all the tests below myself in-game.

## Problem

`AssetManager.Patches.AssetBundleLoader_GetAllAssetPathsMappedToAssetID` calls `SetInstruction` without checking that `MatchForward` found the `Dictionary<string, AssetID>.Add` call. If another mod's transpiler already replaced that call (STUWard 1.3.11+ does, see sighsorry1029/STUWard#2), the match is invalid and `SetInstruction` throws `ArgumentOutOfRangeException`. That breaks `AssetManager`'s static constructor for the rest of the session, so every mod that uses `PrefabManager`/`GUIManager` fails with a `TypeInitializationException`.

## Change

If the match is invalid, log a warning and return the instructions unchanged instead of calling `SetInstruction`. When the match is valid, behavior is identical to before. Skipping is safe when the other mod's replacement is equivalent to `AddSafe`, as STUWard's is (skip null keys, keep the first mapping).

## Testing

Valheim 1.0.15, BepInExPack 5.4.2350, Jotunn 2.30.1 with this patch, STUWard 1.3.15, using this repro plugin: https://github.com/Hampus-Toft/AssetManagerCrashMRE (logs under `logs/`).

| Profile | Original Jotunn | With this change |
| --- | --- | --- |
| Jotunn + STUWard | `ArgumentOutOfRangeException` at `AssetManager.cs:98`, then `TypeInitializationException` | No error, one warning on init, `AssetManager`/`PrefabManager` initialise, `GUIManager.GetSprite` works |
| Jotunn only | Clean, `MatchForward` valid, `AddSafe` applied | Same, no warning |

Not tested: a full session with a Jotunn content mod, other game versions or platforms. No changelog entry added.


sirskunkalot:
This would mean relying on other mods to patch the method correctly, we might need to discuss that internally.

Hampus-Toft:
Correct, but in the current code you do not support external replacing things like this even if its the same replacment but done earlier, as seen in behaviour described.

This "fix" simply makes it not crash but dose not provide any guarantees that it will work as expected.

I belive this should be the responsibility of other mod developers to document incompatibility with Jotunn if they replace functionality that Jotunn relies on.

sirskunkalot:
I am not very happy with the fact that this fix does remove the transpiler and might result in undefined behaviour we have no control of. We are going to reach out to sighsorry and try to find a solution.

sirskunkalot:
Fixed in STUward

